PR #1, opened 2020-07-21
Feature/spine.in.kotlin

soywiz:
Converted Java into Kotlin


soywiz:
Much more work for https://github.com/korlibs/korge/issues/225

PR #5, opened 2020-07-28
Fix box2d samples

RezMike:
Now you don't need to call `.centered` on rect.
I also added a few adjustments, so two samples look the same.


soywiz:
It is not like you don't need it, since since you are synchronizing/generating the body from the view, it should support a view that is already constructed properly with the .centered

PR #17, opened 2020-08-03
Support mouse wheel on mingwX64 (Windows K/N)

RezMike:
Some work on https://github.com/korlibs/korge/issues/639

### DIFF
--- a/korgw/src/mingwX64Main/kotlin/com/soywiz/korgw/DefaultGameWindowMingwX64.kt
+++ b/korgw/src/mingwX64Main/kotlin/com/soywiz/korgw/DefaultGameWindowMingwX64.kt
@@ -1,8 +1,7 @@
 package com.soywiz.korgw
 
-import com.soywiz.kgl.*
-import com.soywiz.klock.DateTime
-import com.soywiz.klock.hr.HRTimeSpan
+import com.soywiz.kgl.toInt
+import com.soywiz.kmem.*
 import com.soywiz.korag.AG
 import com.soywiz.korag.AGConfig
 import com.soywiz.korag.AGOpenglFactory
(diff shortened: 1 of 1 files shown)


korgw/src/mingwX64Main/kotlin/com/soywiz/korgw/DefaultGameWindowMingwX64.kt
```diff
@@ -440,6 +445,11 @@ fun WndProc(hWnd: HWND?, message: UINT, wParam: WPARAM, lParam: LPARAM): LRESULT
             val y = (lParam.toInt() ushr 16) and 0xFFFF
             mouseMove(x, y, wParam.toInt())
         }
+        _WM_MOUSEWHEEL -> {
+            val type = com.soywiz.korev.MouseEvent.Type.SCROLL
+            val scrollDeltaY = ((wParam.toInt() ushr 16) and 0xFFFF).toByte().toDouble()
```
soywiz, line 450: I guess:
```
.extract(16, 8)
```

korgw/src/mingwX64Main/kotlin/com/soywiz/korgw/DefaultGameWindowMingwX64.kt
```diff
@@ -440,6 +445,11 @@ fun WndProc(hWnd: HWND?, message: UINT, wParam: WPARAM, lParam: LPARAM): LRESULT
...
+            windowsGameWindow.mouseEvent(type, mouseX, mouseY, 8, wParam.toInt(), scrollDeltaY)
```
soywiz, line 451: So the values of scrollDeltaY are between -128 and +128? Should be they -1.0, +1.0?

soywiz: Maybe `.convertRange(-128.0, +127.0, -1.0, +1.0)` ?

RezMike: The value is either -128, or 128. I don't know if it's the same on different mingwX64 targets (like windows 7, windows 10, etc.). I think 128 is a number of pixels to scroll

soywiz: https://docs.microsoft.com/en-us/windows/win32/inputdev/wm-mousewheel

> The high-order word indicates the distance the wheel is rotated, expressed in multiples or divisions of WHEEL_DELTA, which is 120. A positive value indicates that the wheel was rotated forward, away from the user; a negative value indicates that the wheel was rotated backward, toward the user.
> The low-order word indicates whether various virtual keys are down. This parameter can be one or more of the following values.

RezMike: Answered in Slack:
"Well, we don't know what values are provided on macos and linux native. So I don't think we should convert values right now. I think we should implement scrolling on all the targets and then check these scroll values and adjust all of them to the same value"

soywiz: It's already implemented on the other targets I guess. At least on linux

PR #23, opened 2020-08-31
Additions to MeshBuilder3D +

dhakehurst:
Add a 'shape3D' to the DSL.
add funs to MeshBuilder3D for cuboid, cube, sphere, rect, triangle
begin support for adding text to 3D shapes.
fix typo shiness --> shininess


soywiz:
Thanks. Very nice contribution. I'll check asap

soywiz:
Thanks for your patience, I had to do lots of chores after the holidays. Checking already

korge/src/commonMain/kotlin/com/soywiz/korge3d/BatchBuilder3D.kt
```diff
@@ -0,0 +1,497 @@
...
+ * Allows to draw quads and sprites buffering the geometry to limit the draw batches executed calling [AG] (Accelerated Graphics).
+ * This class handles a vertex structure of: x, y, u, v, colorMul, colorAdd. Allowing to draw texturized and tinted primitives.
+ *
+ * You should call: [drawQuad], [drawQuadFast], [drawNinePatch], [drawVertices] for buffering the geometries.
+ *
+ * For performance the actual drawing/rendering doesn't happen until the [flush] method is called (normally that happens automatically).
+ * Then the engine will call [flush] when required, automatically once the buffer is filled, at the end of the frame, or when [RenderContext.flush] is executed
+ * by other renderers.
+ */
+class BatchBuilder3D(
```
soywiz, line 28: Why is this class required? Seems a copy of BatchBuilder2D. Since I cannot get the diff, which are the differences, and why is this needed?

korge/src/commonMain/kotlin/com/soywiz/korge3d/Library3D.kt
```diff
@@ -204,7 +204,7 @@ fun Library3D.MaterialDef.instantiate(): Material3D {
 		ambient = effect?.ambient?.instantiate() ?: Material3D.LightColor(Colors.BLACK),
 		diffuse = effect?.diffuse?.instantiate() ?: Material3D.LightColor(Colors.BLACK),
 		specular = effect?.specular?.instantiate() ?: Material3D.LightColor(Colors.BLACK),
-		shiness = effect?.shiness ?: 0.5f,
+        shininess = effect?.shiness ?: 0.5f,
```
soywiz, line 207: I guess you should rename `effect?.shiness` too

korge/src/commonMain/kotlin/com/soywiz/korge3d/BitmapFontExt3D.kt
```diff
@@ -0,0 +1,89 @@
...
+import com.soywiz.korma.geom.Matrix
+import com.soywiz.korma.geom.Vector3D
+import com.soywiz.korma.geom.invert
+
+fun BitmapFont.drawText3D(
+    ctx: RenderContext3D,
+    v1: Vector3D,
+    v2: Vector3D,
+    v3: Vector3D,
+    v4: Vector3D,
```
soywiz, line 21: What is each vector for? Either a small documentation line explaining them, or if there are other names than vN to understand these parameters would be nice

soywiz: As I checked later, this might be the face containing the text? Maybe we can create a class for the Face? Also, could we determine a quad in the 3D space with only three points and compute the other one? Maybe a Face class could allow to construct that instance several ways and provide it here.

korge/src/commonMain/kotlin/com/soywiz/korge3d/MeshBuilder3D.kt
```diff
@@ -1,42 +1,203 @@
...
+import com.soywiz.korim.color.RGBA
+import com.soywiz.korim.font.BitmapFont
+import com.soywiz.korio.util.buildList
+import com.soywiz.korma.geom.Vector3D
+import kotlin.math.PI
+import kotlin.math.cos
+import kotlin.math.sin
 
 @Korge3DExperimental
 class MeshBuilder3D {
```
soywiz, line 18: Nice additions!

korge/src/commonMain/kotlin/com/soywiz/korge3d/Shapes3D.kt
```diff
@@ -1,15 +1,47 @@
...
 
 @Korge3DExperimental
 @Deprecated("Kotlin/Native boxes inline+Number")
-inline fun Container3D.box(width: Number, height: Number, depth: Number, callback: Cube3D.() -> Unit = {}): Cube3D
-    = box(width.toDouble(), height.toDouble(), depth.toDouble(), callback)
+inline fun Container3D.cube(width: Number, height: Number, depth: Number, callback: Cube3D.() -> Unit = {}): Cube3D = cube(width.toDouble(), height.toDouble(), depth.toDouble(), callback)
 
 @Korge3DExperimental
-inline fun Container3D.box(
+inline fun Container3D.cube(
```
soywiz, line 44: We have several terms: `box`, `cube` and `cuboid`. It seems that other programs including blender uses cube so the change makes sense. But in the MeshBuilder3D looks like there is a cuboid method. Is that a normal cube/box, or is something else?

dhakehurst:
no worries

korge/src/commonMain/kotlin/com/soywiz/korge3d/BatchBuilder3D.kt
dhakehurst: I think it is no longer needed.
I was using it as part of the 3D Text impl.
But have had to change the approach.
I did say that the 3DText is still WIP. (maybe it should be on a separate PR)

korge/src/commonMain/kotlin/com/soywiz/korge3d/Library3D.kt
dhakehurst: yes :-)

korge/src/commonMain/kotlin/com/soywiz/korge3d/Shapes3D.kt
dhakehurst: a cube has width == height == depth.
if they are different it is called a cuboid.
(akin to square and rectangle in 2D)

korge/src/commonMain/kotlin/com/soywiz/korge3d/BitmapFontExt3D.kt
dhakehurst: good idea. 
see above comment re 3D Text being in progress

korge/src/commonMain/kotlin/com/soywiz/korge3d/Shapes3D.kt
soywiz: Yaah, that's what I thought. I guess I called `box` (though probably `cuboid` more correct) because of that. But it seems that programs use `cube` for simplicity. Or that initially has all its components to 1,1,1, and you use transforms to deform it converting it into a cuboid.

korge/src/commonMain/kotlin/com/soywiz/korge3d/BatchBuilder3D.kt
soywiz: I see. If this is not required anymore, maybe this can be removed.

Or if you need something very similar to BatchBuilder2D, we can rename BatchBuilder2D to just BatchBuilder, then make that open or abstract so we can reuse the code. Alternatively if we just need some extension methods, maybe the BatchBuilder could be closed, which usually gives better performance on native.

korge/src/commonMain/kotlin/com/soywiz/korge3d/BitmapFontExt3D.kt
soywiz: As you prefer. If you think this PR without the text additions makes sense already, we can merge it, so you can add the text one in a separate PR.

dhakehurst: the 3D Text is proving to be more work than I initially though, so it is probably best to keep that to a diff PR.
If it is easy for you to pick out the shiness typo fixes and MeshBuilder stuff, then do that from this PR.
Else close it and I will try and create a diff branch without the 3D Text bits, and give you a new PR from that.

soywiz: It's Ok. This repo doesn't need to be super stable right now. I can merge as it is right now, so we don't have to rebase stuff

soywiz:
Thanks!

korge/src/commonMain/kotlin/com/soywiz/korge3d/BitmapFontExt3D.kt
dhakehurst: ok. great. I am will continue the 3D Text, and give a separate PR when I have something (more) working.

PR #31, opened 2020-09-19
Moving towards Easings enum class and deprecated most companion object methods

nanodeath:
This is a follow-up to [this post](https://forum.korge.org/topic/30/why-is-korma-s-easing-interface-not-an-enum/2).

There are a couple drawbacks with the current implementation that would be nice to address:
1. Every easing is duplicated between the public `Easing` companion object and the private `Easings` object.
2. It's hard to enumerate through all predefined easings, which is technically possible to do with reflection, but trivial with an enum.

Solving the first item isn't really possible without breaking compatibility, which isn't worth it for this level of minor change IMO.

And this PR does solve the second issue, which I use to write the test I originally wanted.

I'm not 100% sure it's worth merging this change, but I think the

```kotlin
interface Foo {
}
enum class Foos : Foo {
}
```

pattern is a good pattern to follow, in general.

### DIFF
--- a/korma/src/commonMain/kotlin/com/soywiz/korma/interpolation/Easing.kt
+++ b/korma/src/commonMain/kotlin/com/soywiz/korma/interpolation/Easing.kt
@@ -2,92 +2,160 @@ package com.soywiz.korma.interpolation
 
 import kotlin.math.*
 
+private inline fun combine(it: Double, start: Easing, end: Easing) =
+    if (it < 0.5) 0.5 * start(it * 2.0) else 0.5 * end((it - 0.5) * 2.0) + 0.5
+private const val BOUNCE_FACTOR = 1.70158
+private const val HALF_PI = PI / 2.0
+
 @Suppress("unused")
 fun interface Easing {
     operator fun invoke(it: Double): Double
 
     companion object {
         fun cubic(f: (t: Double, b: Double, c: Double, d: Double) -> Double): Easing = Easing { f(it, 0.0, 1.0, 1.0) }
-        fun combine(start: Easing, end: Easing) = Easing { if (it < 0.5) 0.5 * start(it * 2.0) else 0.5 * end((it - 0.5) * 2.0) + 0.5 }
+        fun combine(start: Easing, end: Easing) = Easing { combine(it, start, end) }
 
-        operator fun invoke(f: (Double) -> Double) = Easing { it -> f(it) }
+        operator fun invoke(f: (Double) -> Double) = Easing { f(it) }
 
-        val SMOOTH get() = Easings.SMOOTH
-        val EASE_IN_ELASTIC get() = Easings.EASE_IN_ELASTIC
-        val EASE_OUT_ELASTIC get() = Easings.EASE_OUT_ELASTIC
-        val EASE_OUT_BOUNCE get() = Easings.EASE_OUT_BOUNCE
-        val LINEAR get() = Easings.LINEAR
-        val EASE_IN get() = Easings.EASE_IN
-        val EASE_OUT get() = Easings.EASE_OUT
-        val EASE_IN_OUT get() = Easings.EASE_IN_OUT
-        val EASE_OUT_IN get() = Easings.EASE_OUT_IN
-        val EASE_IN_BACK get() = Easings.EASE_IN_BACK
-        val EASE_OUT_BACK get() = Easings.EASE_OUT_BACK
-        val EASE_IN_OUT_BACK get() = Easings.EASE_IN_OUT_BACK
-        val EASE_OUT_IN_BACK get() = Easings.EASE_OUT_IN_BACK
-        val EASE_IN_OUT_ELASTIC get() = Easings.EASE_IN_OUT_ELASTIC
-        val EASE_OUT_IN_ELASTIC get() = Easings.EASE_OUT_IN_ELASTIC
-        val EASE_IN_BOUNCE get() = Easings.EASE_IN_BOUNCE
-        val EASE_IN_OUT_BOUNCE get() = Easings.EASE_IN_OUT_BOUNCE
-        val EASE_OUT_IN_BOUNCE get() = Easings.EASE_OUT_IN_BOUNCE
-        val EASE_IN_QUAD get() = Easings.EASE_IN_QUAD
-        val EASE_OUT_QUAD get() = Easings.EASE_OUT_QUAD
-        val EASE_IN_OUT_QUAD get() = Easings.EASE_IN_OUT_QUAD
-        val EASE_SINE get() = Easings.EASE_SINE
+        /**
+         * Retrieves a mapping of all standard easings defined directly in [Easing], for example "SMOOTH" -> Easing.SMOOTH.
+         */
+        val ALL: Map<String, Easing> by lazy(LazyThreadSafetyMode.PUBLICATION) {
+            Easings.values().associateBy { it.name }
+        }
+
+        // Author's note:
+        // 1. Make sure new standard easings are added both here and in the Easings enum class
+        // 2. Make sure the name is the same, otherwise ALL_STANDARD will return confusing results
+
+        val SMOOTH: Easing get() = Easings.SMOOTH
+        val EASE_IN_ELASTIC: Easing get() = Easings.EASE_IN_ELASTIC
+        val EASE_OUT_ELASTIC: Easing get() = Easings.EASE_OUT_ELASTIC
+        val EASE_OUT_BOUNCE: Easing get() = Easings.EASE_OUT_BOUNCE
+        val LINEAR: Easing get() = Easings.LINEAR
+        val EASE_IN: Easing get() = Easings.EASE_IN
+        val EASE_OUT: Easing get() = Easings.EASE_OUT
+        val EASE_IN_OUT: Easing get() = Easings.EASE_IN_OUT
+        val EASE_OUT_IN: Easing get() = Easings.EASE_OUT_IN
+        val EASE_IN_BACK: Easing get() = Easings.EASE_IN_BACK
+        val EASE_OUT_BACK: Easing get() = Easings.EASE_OUT_BACK
+        val EASE_IN_OUT_BACK: Easing get() = Easings.EASE_IN_OUT_BACK
+        val EASE_OUT_IN_BACK: Easing get() = Easings.EASE_OUT_IN_BACK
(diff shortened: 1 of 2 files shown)


korma/src/commonMain/kotlin/com/soywiz/korma/interpolation/Easing.kt
```diff
@@ -2,92 +2,171 @@ package com.soywiz.korma.interpolation
...
-        val EASE_OUT_IN_ELASTIC get() = Easings.EASE_OUT_IN_ELASTIC
-        val EASE_IN_BOUNCE get() = Easings.EASE_IN_BOUNCE
-        val EASE_IN_OUT_BOUNCE get() = Easings.EASE_IN_OUT_BOUNCE
-        val EASE_OUT_IN_BOUNCE get() = Easings.EASE_OUT_IN_BOUNCE
-        val EASE_IN_QUAD get() = Easings.EASE_IN_QUAD
-        val EASE_OUT_QUAD get() = Easings.EASE_OUT_QUAD
-        val EASE_IN_OUT_QUAD get() = Easings.EASE_IN_OUT_QUAD
-        val EASE_SINE get() = Easings.EASE_SINE
+        @Deprecated("Moved to Easings enum", ReplaceWith("Easings.SMOOTH"))
+        val SMOOTH: Easing = Easings.SMOOTH
```
soywiz, line 21: Misses `get() =` that might causes issues on initialization order.

In fact, initially I put the implementation here directly without the private class Easings, but had to split it and use getters and failed, not sure if on JS or on Kotlin/Native due to initialization order. For example in JS I getting `unefined` on some cases at runtime.

korma/src/commonMain/kotlin/com/soywiz/korma/interpolation/Easing.kt
```diff
@@ -2,92 +2,171 @@ package com.soywiz.korma.interpolation
...
-    }
-
-    val EASE_OUT_ELASTIC = Easing {
-        if (it == 0.0 || it == 1.0) it else {
-            val p = 0.3
-            val s = p / 4.0
-            2.0.pow(-10.0 * it) * sin((it - s) * (2.0 * PI) / p) + 1
-        }
-    }
+enum class Easings : Easing {
```
soywiz, line 67: What about renaming this to `DefaultEasings` / making it private?

korma/src/commonTest/kotlin/com/soywiz/korma/interpolation/EasingTest.kt
```diff
@@ -26,4 +27,20 @@ class EasingTest {
         assertEquals(0.0, Easing.EASE_SINE(0.0))
         assertEquals(1.0, Easing.EASE_SINE(1.0))
     }
+
```
soywiz, line 30: Nice test additions!

korma/src/commonMain/kotlin/com/soywiz/korma/interpolation/Easing.kt
```diff
@@ -2,92 +2,171 @@ package com.soywiz.korma.interpolation
...
+    SMOOTH {
+        override fun invoke(it: Double): Double = it * it * (3 - 2 * it)
+    },
```
soywiz, line 70: Couldn't be: `SMOOTH(Easing { it * it * (3 - 2 * it) })` ?

korma/src/commonMain/kotlin/com/soywiz/korma/interpolation/Easing.kt
```diff
@@ -2,92 +2,171 @@ package com.soywiz.korma.interpolation
...
+        @Deprecated("Moved to Easings enum", ReplaceWith("Easings.EASE_OUT_IN_BOUNCE"))
+        val EASE_OUT_IN_BOUNCE: Easing get() = Easings.EASE_OUT_IN_BOUNCE
+        @Deprecated("Moved to Easings enum", ReplaceWith("Easings.EASE_IN_QUAD"))
+        val EASE_IN_QUAD: Easing get() = Easings.EASE_IN_QUAD
+        @Deprecated("Moved to Easings enum", ReplaceWith("Easings.EASE_OUT_QUAD"))
+        val EASE_OUT_QUAD: Easing get() = Easings.EASE_OUT_QUAD
+        @Deprecated("Moved to Easings enum", ReplaceWith("Easings.EASE_IN_OUT_QUAD"))
+        val EASE_IN_OUT_QUAD: Easing get() = Easings.EASE_IN_OUT_QUAD
+        @Deprecated("Moved to Easings enum", ReplaceWith("Easings.EASE_SINE"))
+        val EASE_SINE: Easing get() = Easings.EASE_SINE
```
soywiz, line 63: The idea of korge-next is to not have `@Deprecated`.

Initially I had `Easing` and `Easings` as public classes (though Easings was not an enum).
Due to discovery and intuition of the users, I tried to merge them in a single class, but it was not possible due to strange issues on the non-JVM targets.
So in the end I kept two classes, and used getters to get it working.
Having the easings as part of the companion object of Easing, make it easier for people that need an `Easing` instance, to know where to get one.

I think it is a good idea what you proposed to be able to iterate over all the standard provided easings, but I would keep this companion object properties and maybe hiding the enum class. I know it is not fully DRY, but at least is statically typed and the implementation happens only once.

Even if the `Easings` enum is private, you can here put something like:

```
val ALL by lazy { Easings.values() }
// or
val ALL_STANDARD by lazy { Easings.values().associateBy { it.name } }
```

This way we are only exposing `Easing` making it simpler to the users, while also supporting iterating the available standard ones and preventing initialization order issues.

What do you think?

korma/src/commonMain/kotlin/com/soywiz/korma/interpolation/Easing.kt
nanodeath: Actually, I did run into one issue that might be related to this. Initially I didn't have the extra `combine` method and instead stored a `Easing.combine(EASE_IN, EASE_OUT)`-type thing in the Easings that were derived from other Easings. This, confusingly, led to NPE's in the enum constructor, which _might_ be relaetd to what you're talking about.

> might 

Would existing tests catch these?

korma/src/commonMain/kotlin/com/soywiz/korma/interpolation/Easing.kt
nanodeath: I definitely want it to be public...IMO that's half the point of this change. I like the name `Easings` because it's more discoverable, but it does have the problem that people might be tempted to use the `Easings` type directly in their code instead of the interface. If it was named `DefaultEasings` that would be more noticeably an antipattern to people :)

You okay with naming all future interface-enum enums as Default*?

korma/src/commonMain/kotlin/com/soywiz/korma/interpolation/Easing.kt
nanodeath: You mean with enum having a constructor that stores an Easing? I guess, but it seems like (IMO, too much) overhead for just making the syntax more compact.

korma/src/commonMain/kotlin/com/soywiz/korma/interpolation/Easing.kt
nanodeath: I definitely agree that `korge-next`, because it's a new major version, shouldn't have deprecated stuff; that should go in the 1.x version, then people will see them before they upgrade to 2.x.

What you suggest sounds reasonable; I'll make that change. And I like the `ALL_STANDARD` variant.

nanodeath:
Thanks for looking!

nanodeath:
I've incorporated most of your feedback into a new revision. Enum class is now private, nothing's deprecated, however I still think it makes sense to just define the easing itself as an implementation of the method interface. I like the idea of making `Easing.ALL` lazy, so I did that but included `LazyThreadSafetyMode.PUBLICATION` -- should be a little more efficient without sacrificing anything.

korma/src/commonMain/kotlin/com/soywiz/korma/interpolation/Easing.kt
```diff
@@ -2,92 +2,160 @@ package com.soywiz.korma.interpolation
...
-        val EASE_IN_BOUNCE get() = Easings.EASE_IN_BOUNCE
-        val EASE_IN_OUT_BOUNCE get() = Easings.EASE_IN_OUT_BOUNCE
-        val EASE_OUT_IN_BOUNCE get() = Easings.EASE_OUT_IN_BOUNCE
-        val EASE_IN_QUAD get() = Easings.EASE_IN_QUAD
-        val EASE_OUT_QUAD get() = Easings.EASE_OUT_QUAD
-        val EASE_IN_OUT_QUAD get() = Easings.EASE_IN_OUT_QUAD
-        val EASE_SINE get() = Easings.EASE_SINE
-    }
-}
+        operator fun invoke(f: (Double) -> Double) = Easing { f(it) }
```
soywiz, line 18: This method should not be required anymore since changed `interface Easing` into `fun interface Easing`, which is why I suggested to change enum members to `SOMETHING(Easing { })`. Please check this: https://kotlinlang.org/docs/reference/whatsnew14.html#sam-conversions-for-kotlin-interfaces

korma/src/commonMain/kotlin/com/soywiz/korma/interpolation/Easing.kt
soywiz: Explained here: https://github.com/korlibs/korge-next/pull/31/files#r491484527 (might require removing the old invoke method)

korma/src/commonMain/kotlin/com/soywiz/korma/interpolation/Easing.kt
soywiz: I'm not 100% sure, but I think initialization order is something that might cause issues not always but in some cases. When I wrote JTransc and tried to do static initialization found some issues regarding to cycles in the objects that depending on the tree shaking / dead code elimination (classes and methods included) caused issues sometimes and was avoided others by chance.

In the case of Easing I remember having those problems but can't tell you the details because I simply don't remember them

soywiz (changes requested):
Nice changes.
~Can you check the SAM conversion for kotlin interfaces from 1.4? Maybe you can remove the `invoke` method and make the implementations one-liners?
After changing that I think this is a great addition, thanks!~

Edited

soywiz:
Sorry, reviewed without a computer. I now understand that Easings enum implements the interface and thus is not requiring a lambda as argumented and that what I suggested was not possible. Merged already. Thanks!

PR #35, opened 2020-09-28
Skybox, etc

dhakehurst:
+ progress on 3D
 - skybox
 - terrain (could do with more work)
 - text3D (needs more work)
 - Camera Movement functions and switch to "left hand rule" for 3D axis (openGL standard)
 - 3D mesh builder uses indices

+ Other
 - add facility to switch on logging to Korge
 - arg to allow Korge to run without blocking thread
 - other mods to enable skybox


korge/src/commonMain/kotlin/com/soywiz/korge3d/tween/tween.kt
```diff
@@ -0,0 +1,304 @@
...
+import kotlin.reflect.*
+
+class TweenComponent(
+        override val view: View3D,
+        private val vs: List<V2<*>>,
+        val time: HRTimeSpan = HRTimeSpan.NIL,
+        val easing: Easing = DEFAULT_EASING,
+        val callback: (Double) -> Unit,
+        val c: CancellableContinuation<Unit>
+) : UpdateComponentV2 {
```
soywiz, line 27: This whole file is duplicated. I think we should try to figure out a way to deduplicate it. Maybe we can create something like BaseView interface that is implemented by View and View3D? Or extract the functionality in a place that could be called in both cases

soywiz (approved):

soywiz:
Thanks @dhakehurst ! A really nice contribution.
I will try to generalize some duplicated code so we can deduplicate it, I will try to do that in a PR so you can review it

dhakehurst:
Yes I started copying the tween stuff to 3d. But then stopped as i couldn’t
see how it was supposed to be working. And i also though it was silly to
have a duplicate. Perhaps a generic arg could help?

Am unsure if we can make it common. 3d of course needs an additional
dimension!

soywiz:
In the end I have deduplicated it like this (created BaseView): https://github.com/korlibs/korge-next/commit/cf4baf3c003b08d5f7318eedff45acdfb3074596

dhakehurst:
looks like it could work...would probably have to try it out to be sure.
good start.

PR #45, opened 2020-10-08
Unit test for bone transformation

tradfursten:


korge-spine/src/commonTest/kotlin/com/esotericsoftware/spine/BoneTest.kt
```diff
@@ -0,0 +1,41 @@
...
+import com.soywiz.korma.geom.Vector3D
+import kotlin.math.absoluteValue
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertTrue
+import kotlin.test.asserter
+
+class BoneTest {
+    @Test
+    fun testSimpleRotation() {
```
soywiz, line 15: Nice. Thanks!

korge-spine/src/commonTest/kotlin/com/esotericsoftware/spine/BonePositionTest.kt
```diff
@@ -0,0 +1,57 @@
...
+        val skeletonData = json.readSkeletonData(resource)
+        val skeleton = Skeleton(skeletonData)
+        val bone = skeleton.findBone("gun-tip")
+        val fps = 1 / 15f
+        for (animation in skeletonData.animations) {
+            var time = 0f
+            while (time < animation.duration) {
+                animation.apply(skeleton, time, time, false, null, 1f, MixBlend.first, MixDirection.`in`)
+                skeleton.updateWorldTransform()
+                println(animation.name + "," + bone!!.worldX + "," + bone.worldY + "," + bone.worldRotationX)
```
soywiz, line 26: Instead of printing, could you emit the values and then check/assert them against a reference?

korge-spine/src/commonTest/kotlin/com/esotericsoftware/spine/BonePositionTest.kt
```diff
@@ -0,0 +1,57 @@
...
+                time += fps
+            }
+        }
+    }
+}
+
+class AttachmentLoaderTest: AttachmentLoader {
```
soywiz, line 33: To not require this class, maybe you could go to the AttachmentLoader interface and mark all the methods with a default implementation, or even making it an open class:

```
interface AttachmentLoader {
    fun newRegionAttachment(skin: Skin, name: String, path: String): RegionAttachment?
    // ...
}
```
->
```
interface AttachmentLoader {
    fun newRegionAttachment(skin: Skin, name: String, path: String): RegionAttachment? = null
    // ...
}
```
or
```
open class AttachmentLoader {
    open fun newRegionAttachment(skin: Skin, name: String, path: String): RegionAttachment? = null
    // ...
}
```

After that,
```
val json = SkeletonJson(AttachmentLoaderTest())
```
->
```
val json = SkeletonJson(object : AttachmentLoader { }) // interface version
val json = SkeletonJson(object : AttachmentLoader() { }) // class version
```

tradfursten: I'll go for the interface version

korge-spine/src/commonTest/kotlin/com/esotericsoftware/spine/BonePositionTest.kt
tradfursten: This test was just a test imported from another spine runtime :)
Is it ok if I create a reference from one execution and then validate against that so we'll catch errors if things change ?

soywiz: Maybe we can stringify float/doubles by just taking one decimal into account. For example: `10.0.toStringDecimal(1)`
I guess that might prevent any rounding errors. Then you can make a single assert with a string with all the values already on it, so it is easier to change later.
You can also use:
```
"""
data...
""".trimIndent()
```
to store the values as a string, so it is easy to copy and paste if the assert changes. You can check a sample of this here:
https://github.com/korlibs/korge-next/blob/d757f79c057d0c2480fc665703cf2383ba011560/korge/src/jvmTest/kotlin/com/soywiz/korge/bitmapfont/BitmapFontTest.kt#L30-L68

Unless the output is going to be stable, I usually like to have as few tests like that as possible. Maybe one per entity.
But here I guess that if the matrix output is consistent, this bone test should be stable enough.

korge-spine/src/commonTest/kotlin/com/esotericsoftware/spine/BonePositionTest.kt
```diff
@@ -7,51 +7,31 @@ import com.soywiz.korio.async.suspendTest
...
         val fps = 1 / 15f
+        var line = 0
         for (animation in skeletonData.animations) {
             var time = 0f
             while (time < animation.duration) {
                 animation.apply(skeleton, time, time, false, null, 1f, MixBlend.first, MixDirection.`in`)
                 skeleton.updateWorldTransform()
-                println(animation.name + "," + bone!!.worldX + "," + bone.worldY + "," + bone.worldRotationX)
+                assertEquals(testResult[line++],"${animation.name},${bone!!.worldX},${bone.worldY},${bone.worldRotationX}")
+                println("${animation.name},${bone!!.worldX},${bone.worldY},${bone.worldRotationX}")
```
soywiz, line 31: The problem with this approach is that if for example if 10 output lines change, the way to update them is way harder  than with a single assert, also even when running fine you are outputing 155 lines.

With a single assert you don't need the print, when using the IDE comparing the strings you can just copy and paste. 
Initially I had problems with resources in tests on multiplatform, specially on JS. Not sure if that's fixed. I usually embed strings because it is easier to find the code and the expectation, but if it works in a separate file, I'm ok with it.

a
b

tradfursten: I didn't know about that feature in intelliJ, thanks for the tip. I'll change to a test of one of the animations and assert agains a string instead of a file content

soywiz (approved):

soywiz:
Thank you. Nice PR!

PR #76, opened 2020-11-15
This should fix the transposed stick axes on gamepads.

stewsters:
The input from gamepads switched some of the axes.  This should fix it.


soywiz:
Nice, thanks!

PR #77, opened 2020-11-15
3D double to float

KenVanHoeylandt:
As discussed on Discord, this converts `double` to `float`.
This still needs some more testing, but Big Sur is giving me serious performance issues with Android Studio and related tooling.

Besides the `double` to `float` conversion, I took the liberty to align the naming a bit more (e.g. `setToFunc` -> `setTo`).

### DIFF
--- a/korge/src/commonMain/kotlin/com/soywiz/korge3d/RenderContext3D.kt
+++ b/korge/src/commonMain/kotlin/com/soywiz/korge3d/RenderContext3D.kt
@@ -6,7 +6,7 @@ import com.soywiz.korge.render.*
 import com.soywiz.korma.geom.*
 
 @Korge3DExperimental
-class RenderContext3D() {
+class RenderContext3D {
 	lateinit var ag: AG
 	lateinit var rctx: RenderContext
 	val shaders = Shaders3D()
@@ -21,6 +21,5 @@ class RenderContext3D() {
 	val cameraMatInv: Matrix3D = Matrix3D()
 	val dynamicVertexBufferPool = Pool { ag.createVertexBuffer() }
     val dynamicIndexBufferPool = Pool { ag.createIndexBuffer() }
-	val ambientColor: Vector3D = Vector3D()
-
+	val ambientColor: Vector3D = Vector3D() // TODO: This shouldn't be a Vector3D
 }
--- a/korma/src/commonMain/kotlin/com/soywiz/korma/geom/Quaternion.kt
+++ b/korma/src/commonMain/kotlin/com/soywiz/korma/geom/Quaternion.kt
@@ -2,80 +2,85 @@ package com.soywiz.korma.geom
 
 import kotlin.math.*
 
+inline fun quaternionLength(x: Float, y: Float, z: Float, w: Float) = sqrt(quaternionLengthSquared(x, y, z, w))
+inline fun quaternionLengthSquared(x: Float, y: Float, z: Float, w: Float) = (x * x) + (y * y) + (z * z) + (w * w)
+
 // https://en.wikipedia.org/wiki/Conversion_between_quaternions_and_Euler_angles
 data class Quaternion(
-    var x: Double = 0.0,
-    var y: Double = 0.0,
-    var z: Double = 0.0,
-    var w: Double = 1.0
+    var x: Float = 0f,
+    var y: Float = 0f,
+    var z: Float = 0f,
+    var w: Float = 1f
 ) {
     companion object {
-        fun dotProduct(l: Quaternion, r: Quaternion): Double = l.x * r.x + l.y * r.y + l.z * r.z + l.w * r.w
-        operator fun invoke(x: Float, y: Float, z: Float, w: Float) = Quaternion(x.toDouble(), y.toDouble(), z.toDouble(), w.toDouble())
-        operator fun invoke(x: Int, y: Int, z: Int, w: Int) = Quaternion(x.toDouble(), y.toDouble(), z.toDouble(), w.toDouble())
+        fun dotProduct(l: Quaternion, r: Quaternion): Float = l.x * r.x + l.y * r.y + l.z * r.z + l.w * r.w
+
         fun toEuler(q: Quaternion, out: EulerRotation = EulerRotation()): EulerRotation = toEuler(q.x, q.y, q.z, q.w, out)
-        fun toEuler(x: Double, y: Double, z: Double, w: Double, euler: EulerRotation = EulerRotation()): EulerRotation =
-            toEuler(x.toFloat(), y.toFloat(), z.toFloat(), w.toFloat(), euler)
-        fun toEuler(x: Int, y: Int, z: Int, w: Int, euler: EulerRotation = EulerRotation()): EulerRotation =
-            toEuler(x.toFloat(), y.toFloat(), z.toFloat(), w.toFloat(), euler)
 
         fun toEuler(x: Float, y: Float, z: Float, w: Float, out: EulerRotation = EulerRotation()): EulerRotation {
-            val sinrCosp = +2.0 * (w * x + y * z)
-            val cosrCosp = +1.0 - 2.0 * (x * x + y * y)
+            val sinrCosp = +2f * (w * x + y * z)
+            val cosrCosp = +1f - 2f * (x * x + y * y)
             val roll = atan2(sinrCosp, cosrCosp)
-            val sinp = +2.0 * (w * y - z * x)
+            val sinp = +2f * (w * y - z * x)
             val pitch = when {
-                abs(sinp) >= 1 -> if (sinp > 0) PI / 2 else -PI / 2
+                abs(sinp) >= 1f -> if (sinp > 0f) PI.toFloat() / 2f else -PI.toFloat() / 2f
                 else -> asin(sinp)
             }
-            val sinyCosp = +2.0 * (w * z + x * y)
-            val cosyCosp = +1.0 - 2.0 * (y * y + z * z)
+            val sinyCosp = +2f * (w * z + x * y)
+            val cosyCosp = +1f - 2f * (y * y + z * z)
             val yaw = atan2(sinyCosp, cosyCosp)
-            return out.setTo(roll.radians, pitch.radians, yaw.radians)
+            out.setTo(roll.radians, pitch.radians, yaw.radians)
+            return out
         }
     }
 
-    operator fun get(index: Int): Double = when (index) {
+    fun copy() = Quaternion(x, y, z, w)
+
+    inline val length: Float get() = sqrt(quaternionLengthSquared(x, y, z, w))
+    inline val lengthSqr: Float get() = quaternionLengthSquared(x, y, z, w)
+
+    operator fun get(index: Int): Float = when (index) {
         0 -> x
         1 -> y
         2 -> z
         3 -> w
-        else -> Double.NaN
+        else -> Float.NaN
     }
-    inline fun setToFunc(callback: (Int) -> Double) = setTo(callback(0), callback(1), callback(2), callback(3))
-    fun setTo(x: Double, y: Double, z: Double, w: Double): Quaternion {
+    inline fun setToFunc(callback: (Int) -> Float) = setTo(callback(0), callback(1), callback(2), callback(3))
+
+    fun setTo(x: Float, y: Float, z: Float, w: Float): Quaternion {
         this.x = x
         this.y = y
         this.z = z
         this.w = w
         return this
     }
-    fun setTo(x: Int, y: Int, z: Int, w: Int): Quaternion = setTo(x.toDouble(), y.toDouble(), z.toDouble(), w.toDouble())
-    fun setTo(x: Float, y: Float, z: Float, w: Float): Quaternion = setTo(x.toDouble(), y.toDouble(), z.toDouble(), w.toDouble())
-    fun setTo(euler: EulerRotation): Quaternion = EulerRotation.toQuaternion(euler, this)
-    fun setTo(other: Quaternion): Quaternion = setTo(other.x, other.y, other.z, other.w)
+
+    fun setTo(euler: EulerRotation) = setTo(EulerRotation.toQuaternion(euler, this))
+    fun setTo(other: Quaternion) = setTo(other.x, other.y, other.z, other.w)
 
     fun setEuler(x: Angle, y: Angle, z: Angle): Quaternion = EulerRotation.toQuaternion(x, y, z, this)
     fun setEuler(euler: EulerRotation): Quaternion = EulerRotation.toQuaternion(euler, this)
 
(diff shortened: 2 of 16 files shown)


korge/src/commonMain/kotlin/com/soywiz/korge3d/RenderContext3D.kt
```diff
@@ -21,6 +21,5 @@ class RenderContext3D() {
 	val cameraMatInv: Matrix3D = Matrix3D()
 	val dynamicVertexBufferPool = Pool { ag.createVertexBuffer() }
     val dynamicIndexBufferPool = Pool { ag.createIndexBuffer() }
-	val ambientColor: Vector3D = Vector3D()
-
+	val ambientColor: Vector3D = Vector3D() // TODO: This shouldn't be a Vector3D
```
soywiz, line 24: It could be either:
https://github.com/korlibs/korim/blob/2dda8bacac23234a5429f860bbb447ae2f1ad134/korim/src/commonMain/kotlin/com/soywiz/korim/color/RGBAf.kt
Or:
https://github.com/korlibs/korim/blob/2dda8bacac23234a5429f860bbb447ae2f1ad134/korim/src/commonMain/kotlin/com/soywiz/korim/color/RGBA.kt

In 2D graphics, we are using multiplicative tinting with `colorMul: RGBA` and additive tinting with: `colorAdd: ColorAdd` both are packed as Int32, to the ColorAdd to support `[-255, +255]` range loses a bit of precission

korma/src/commonMain/kotlin/com/soywiz/korma/geom/Vector3D.kt
```diff
@@ -5,95 +5,190 @@ import com.soywiz.korma.interpolation.interpolate
...
-    var w: Float get() = data[3]; set(value) = run { data[3] = value }
 
-    val lengthSquared: Float get() = (x * x) + (y * y) + (z * z) + (w * w)
+    val lengthSquared: Float get() = (x * x) + (y * y) + (z * z)
     val length: Float get() = sqrt(lengthSquared)
 
-    val length3Squared: Float get() = (x * x) + (y * y) + (z * z)
-    val length3: Float get() = sqrt(length3Squared)
+    constructor(x: Float, y: Float, z: Float): this(floatArrayOf(x, y, z))
+    constructor(func: (index: Int) -> Float): this(func(0), func(1), func(2))
```
soywiz, line 22: This could be (to prevent allocating the lambda):

```
companion object {
    inline operator fun invoke(func: (index: Int) -> Float) = FloatArray(4) { func(it) }
}
```

korma/src/commonMain/kotlin/com/soywiz/korma/geom/Vector3D.kt
```diff
@@ -5,95 +5,190 @@ import com.soywiz.korma.interpolation.interpolate
...
+inline fun vectorLength(x: Float, y: Float, z: Float) = sqrt(vectorLengthSquared(x, y, z))
+inline fun vectorLengthSquared(x: Float, y: Float, z: Float) = (x * x) + (y * y) + (z * z)
 
+class Vector3D constructor(
+    val data: FloatArray
+) {
     var x: Float get() = data[0]; set(value) = run { data[0] = value }
     var y: Float get() = data[1]; set(value) = run { data[1] = value }
     var z: Float get() = data[2]; set(value) = run { data[2] = value }
-    var w: Float get() = data[3]; set(value) = run { data[3] = value }
```
soywiz, line 14: This was not voted. Removing `w`, would require providing Vector4 and Vector3.

The idea is that Vector3D both supports Vector3 and Vector4 with a limited API. As commented, mimics the AS3 API:
https://help.adobe.com/en_US/FlashPlatform/reference/actionscript/3/flash/geom/Vector3D.html

If you plan to propose this, please keep the Vector3D maybe as deprecated, and introduce two new classes: Vector3 and Vector4. Or at least just Vector3, but keeping Vector3D.

korma/src/commonMain/kotlin/com/soywiz/korma/geom/Quaternion.kt
```diff
@@ -84,54 +89,71 @@ data class Quaternion(
...
+        val originalLength = length
+        setTo(v.x / originalLength, v.y / originalLength, v.z / originalLength, v.w / originalLength)
+        return originalLength
     }
-    fun toMatrix(out: Matrix3D = Matrix3D()): Matrix3D = out.multiply(
+
+    /**
+     * Normalize, except when the length is 0
+     */
+    fun normalizeSafely(v: Quaternion = this): Float {
```
soywiz, line 148: Why don't make `normalize` do this instead? Is this the typical expected behaviour? Maybe normalize can do this, and if you believe some people would want to have NaNs/Infinites here, we can also provide `normalizeUnsafe`

korma/src/commonMain/kotlin/com/soywiz/korma/geom/Quaternion.kt
```diff
@@ -2,80 +2,85 @@ package com.soywiz.korma.geom
...
         func(l.x, r.x),
         func(l.y, r.y),
         func(l.z, r.z),
         func(l.w, r.w)
     )
-    fun setToSlerp(left: Quaternion, right: Quaternion, t: Double, tleft: Quaternion = Quaternion(), tright: Quaternion = Quaternion()): Quaternion {
-        val tleft = tleft.copyFrom(left).normalize()
-        val tright = tright.copyFrom(right).normalize()
+
+    fun setToSlerp(left: Quaternion, right: Quaternion, t: Float, tleft: Quaternion = Quaternion(), tright: Quaternion = Quaternion()) {
```
soywiz, line 79: Did we vote in the end removing chainable calls? If not, we can make a RFC now and vote

korma/src/commonMain/kotlin/com/soywiz/korma/geom/Vector3D.kt
```diff
@@ -5,95 +5,190 @@ import com.soywiz.korma.interpolation.interpolate
...
-    fun scale(scale: Double) = scale(scale.toFloat())
+    fun scale(scale: Float) = setTo(x * scale, y * scale, z * scale)
 
     fun transform(mat: Matrix3D) = mat.transform(this, this)
 
-    fun normalize(vector: Vector3D = this): Vector3D = this.apply {
-        val norm = 1.0 / vector.length3
-        setTo(vector.x * norm, vector.y * norm, vector.z * norm, 1.0)
+    /**
+     * Normalizes this object and returns the length before normalization.
```
soywiz, line 58: Do you have other APIs as reference doing this to determine if this is a common practice?
With chainable, you could do: `normalize().length` that better states what you are doing.
And if this is an optimization, maybe a method called `normalizeAndReturnLength()` would state better the intent of this function

korma/src/commonMain/kotlin/com/soywiz/korma/geom/Vector3D.kt
```diff
@@ -5,95 +5,190 @@ import com.soywiz.korma.interpolation.interpolate
...
+    fun normalized(): Vector3D {
+        return this / length
     }
 
-    fun normalized(out: Vector3D = Vector3D()): Vector3D = out.copyFrom(this).normalize()
+    /**
+     * Normalizes this object and returns the length before normalization.
+     * If the vector length is 0, the vector is set to (0, 0, 1.0f).
+     */
+    fun normalizeSafely(): Float {
```
soywiz, line 78: As stated before, why not use the plain word `normalize` for the safe one, and then an unsafe one to scratch further performance or work with nans?

korma/src/commonMain/kotlin/com/soywiz/korma/geom/Vector3D.kt
```diff
@@ -5,95 +5,190 @@ import com.soywiz.korma.interpolation.interpolate
...
-        (a.x * b.y - a.y * b.x),
-        1f
+    fun setToCross(l: Vector3D, r: Vector3D) = setTo(
+        (l.y * r.z - l.z * r.y),
+        (l.z * r.x - l.x * r.z),
+        (l.x * r.y - l.y * r.x)
     )
-    override fun equals(other: Any?): Boolean = (other is Vector3D) && almostEquals(this.x, other.x) && almostEquals(this.y, other.y) && almostEquals(this.z, other.z) && almostEquals(this.w, other.w)
+
+    // region data class / JVM
```
soywiz, line 100: Actually all the Kotlin targets uses these methods, so not related to JVM

korma/src/commonMain/kotlin/com/soywiz/korma/geom/Vector3D.kt
```diff
@@ -5,95 +5,190 @@ import com.soywiz.korma.interpolation.interpolate
...
+
+    override fun equals(other: Any?): Boolean = (other is Vector3D) && almostEquals(this.x, other.x) && almostEquals(this.y, other.y) && almostEquals(this.z, other.z)
     override fun hashCode(): Int = data.contentHashCode()
+    override fun toString(): String = "(${x.niceStr}, ${y.niceStr}, ${z.niceStr})"
+
+    // endregion
+
+    // region Operator overloading
```
soywiz, line 108: 👍

korma/src/commonMain/kotlin/com/soywiz/korma/geom/Vector3D.kt
```diff
@@ -5,95 +5,190 @@ import com.soywiz.korma.interpolation.interpolate
...
+        x /= other
+        y /= other
+        z /= other
+    }
 
-    override fun toString(): String = if (w == 1f) "(${x.niceStr}, ${y.niceStr}, ${z.niceStr})" else "(${x.niceStr}, ${y.niceStr}, ${z.niceStr}, ${w.niceStr})"
+    // endregion
 }
 
 inline class IntVector3(val v: Vector3D) {
```
soywiz, line 185: This API was inconsisent as I can see now. Since it had 4 components but was called 3 instead of 4 or 3D.

korma/src/commonMain/kotlin/com/soywiz/korma/geom/Vector3D.kt
```diff
@@ -5,95 +5,190 @@ import com.soywiz.korma.interpolation.interpolate
...
     val x: Int get() = v.x.toInt()
     val y: Int get() = v.y.toInt()
     val z: Int get() = v.z.toInt()
-    val w: Int get() = v.w.toInt()
 }
 
-fun Vector3D.asIntVector3D() = IntVector3(this)
+fun Vector3D.toIntvector() = IntVector3(this)
```
soywiz, line 191: Missing case on the `v`
`.toIntvector`
->
`.toIntVector`

Also, the `as` means something. Let me explain the rationale:
* `to` was used when performing a conversion, potentially costly
* `as` was used when inline-wrapping another class, to state that there is like a reinterpretation with no conversion cost

korma/src/commonMain/kotlin/com/soywiz/korma/interpolation/Interpolation.kt
```diff
@@ -8,8 +8,12 @@ interface MutableInterpolable<T> {
     fun setToInterpolated(ratio: Double, l: T, r: T): T
 }
 
-fun Double.interpolate(l: Float, r: Float): Float = (l + (r - l) * this).toFloat()
+fun Float.interpolate(l: Float, r: Float): Float = (l + (r - l) * this)
```
soywiz, line 11: Can you add `Double`, `Int` and `Long` too? The receiver is the ratio, so would make sense to be able to interpolate primitives here

korma/src/commonTest/kotlin/com/soywiz/korma/geom/Matrix3DTest.kt
```diff
@@ -100,32 +99,32 @@ class Matrix3DTest {
             val mat = Matrix(2, 0, 0, 2, 20, 20)
             val mat4 = mat.toMatrix3D()
             assertEquals(Point(40, 40), mat.transform(Point(10, 10)))
-            assertEquals(Vector3D(40, 40, 0), mat4.transform(Vector3D(10f, 10f, 0f)))
+            assertEquals(Vector3D(40f, 40f, 0f), mat4.transform(Vector3D(10f, 10f, 0f)))
```
soywiz, line 102: Didn't notice, the constructor receiving Ints was removed?

Hopefully at some point this won't be necessary if Kotlin implements something like this:
https://youtrack.jetbrains.com/issue/KT-39136

korma/src/commonTest/kotlin/com/soywiz/korma/geom/Matrix3DTest.kt
```diff
@@ -100,32 +99,32 @@ class Matrix3DTest {
...
+            val projection = Matrix3D().apply { setToOrtho(0f, 200f, 100f, 0f, 0f, -20f) }
+            assertEquals(Vector3D(0f, 0f, -1f), Vector3D(100f, 50f, 0f).transform(projection))
+            assertEquals(Vector3D(0f, 0f, +1f), Vector3D(100f, 50f, 20f).transform(projection))
         }
         run {
-            val projection = Matrix3D().setToOrtho(0f, 200f, 100f, 0f, 0f, +20f)
-            assertEquals(Vector3D(0, 0, -1), Vector3D(100f, 50f, 0f).transform(projection))
-            assertEquals(Vector3D(0, 0, +1), Vector3D(100f, 50f, -20f).transform(projection))
+            val projection = Matrix3D().apply { setToOrtho(0f, 200f, 100f, 0f, 0f, +20f) }
+            assertEquals(Vector3D(0f, 0f, -1f), Vector3D(100f, 50f, 0f).apply { transform(projection) })
```
soywiz, line 121: For me, the extra apply looks not too good. Let's do the chainable RFC.

soywiz (commented):
Thanks for the proposals! I thought this PR was going to be just for changing `Double` to `Float`, that was mostly good for me for the 3D API despite of the potentially inconsistencies on the JS target by not being in a FloatArray.
But I believe the other changes should be discussed independently and as result of a consensus as we discussed on Discord already.
Mostly dropping chainable support that is widely used on the library and dropping the w component on Vector3D.

If you can make a PR changing Double -> Float, we can merge it already 👍 .

Then in separate PRs, so they are different battles, you can propose:
* `Vector3D.w` component removal
* The Safely method additions, that maybe could be `plain` +`plainUnsafe` 
* Method renaming include to -> as etc.
* Removing chainable support. I think we can make a poll or something if you are really willing to propose this change

soywiz:
Hey, going to close this since it has been a while already.
The main problem here was that in JS there are no real floats, so this might introduce inconsistencies unless backed from an array, and even in that case.
At some point Kotlin will provide a WASM target, and korge will probably focus on that, and at that point we will be able to use floats for sure.
Also when value objects with multiple fields are supported, this code will change so the geometries are immutable so this will have even more conflicts.

Thanks for your contribution!

PR #107, opened 2021-01-06
Samples - Improve Asteroids and Vectors

Inego:


soywiz:
removeChildren

Thanks!

PR #109, opened 2021-01-09
Rectangle corners

SerVB:
Hi! I think one change from korlibs/korma#47 should be definitely presented in the repo: corner functions for RectangleInt, as Rectangle already has these functions.
The second change in this PR is moving corner functions from Rectangle to IRectangle, to make them available for immutable version of objects.


soywiz:
4t6pdw

Thanks!

PR #113, opened 2021-01-11
fixes android.Context memory leak when using resourcesVfs on android …

pablocaviglia-baires:
…target


soywiz:
leak





Awesome Pablo! Thanks for making this PR during your holidays :)

We should rethink this stuff on 3.0 so it is not static somehow, so you can actually enjoy in a beach and be able to code for fun, not for leaks. 😂

pablocaviglia-baires:
Jajaja, thanks my friend. Actually you made my holidays with this, it kept me thinking while drinking a beer :) 
And yes, we can have a talk later to review this on 3. The good thing is that we have learned lots of new stuff with this!!

soywiz:
Yeah, as we discussed, resourcesVfs is global, but still to access resources on Android we need the context. So let's see if we can figure something that works fine in 3.0. For now I don't have any better ideas

PR #115, opened 2021-01-12
fixes leak on vfs android, fixes leak on korvi androd

pablocaviglia-baires:
fixes leak on vfs android, fixes leak on korvi androd


soywiz:
Thanks!

PR #116, opened 2021-01-12
Makes Logger thread safe

Mayankmkh:
The program would crash when accessing Logger from multiple threads in K/N due to the memory model.
As discussed on Slack https://kotlinlang.slack.com/archives/CJEF0LB6Y/p1609680019381100, the changes make Logger thread-safe by using atomic reference.

### DIFF
--- a/klogger/build.gradle
+++ b/klogger/build.gradle
@@ -1,7 +1,1 @@
-repositories {
-    mavenCentral()
-}
-
-dependencies {
-    commonMainApi("org.jetbrains.kotlinx:atomicfu:0.15.1")
-}
+apply plugin: 'kotlinx-atomicfu'
--- a/klogger/build.gradle
+++ b/klogger/build.gradle
@@ -1,0 +1,7 @@
+repositories {
+    mavenCentral()
+}
+
+dependencies {
+    commonMainApi("org.jetbrains.kotlinx:atomicfu:0.15.1")
+}
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -18,6 +18,7 @@ buildscript {
         classpath("com.android.tools.build:gradle:4.0.1")
         //classpath("com.android.tools.build:gradle:4.1.0-rc03")
         //classpath("com.android.tools.build:gradle:4.2.0-alpha12")
+        classpath("org.jetbrains.kotlinx:atomicfu-gradle-plugin:0.15.1")
     }
 }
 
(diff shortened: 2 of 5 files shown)


klogger/build.gradle
```diff
@@ -0,0 +1,7 @@
+repositories {
+    mavenCentral()
+}
+
+dependencies {
+    commonMainApi("org.jetbrains.kotlinx:atomicfu:0.15.1")
```
soywiz, line 6: The last time I checked, atomicfu required a plugin to be applied? They have lifted that limitation already?

Mayankmkh: Yes, you are right. It does require a plugin. I was trying to add the dependency but Gradle sync was failing so I tried many approaches and this worked. Let me try again to use the plugin instead.
It's weird that the code is working fine without plugin too 🤔

Mayankmkh: @soywiz the plugin is working now but the older version is required (0.14.4 instead of 0.15.1). Some minor changes will be required as the atomic delegation was introduced in 0.15
The build fails using the latest version as mentioned in Kotlin/kotlinx.atomicfu#171. As mentioned in the issue comments, updating the gradle version to 6.8 works fine

Mayankmkh: Should I update Gradle to 6.8 or use older version of the library?

soywiz: Bumped gradle here: https://github.com/korlibs/korge-next/commit/3a3c76bcf27686a00f83a6c1f8f77e6dcc0e72a4

Is the plugin required in the end? Would that force to apply the plugin to all the libraries and applications using klogger, or it is just required for the library directly depending on it?
Sorry if the question is dummy, but since never used it, I really don't know.

Mayankmkh: It is required by KLogger as the plugin does some bytecode transformations for eg optimizing atomic delegates. It is recommended by the AtomicFu lib to

> Declare atomic variables as private val or internal val. You can use just (public) val, but make sure they are not directly accessed outside of your Kotlin module (outside of the source set). Access to the atomic variable itself shall be encapsulated.

It should not be a problem if we follow this. Users of klogger need not apply plugin from what I can infer as I tried running samples with logging enabled and they were working fine.

Thanks for updating the gradle version. I'll make changes to PR.

soywiz: Nice. When synchronizing to the klogger repo I have to remember to apply the plugin and upgrade the gradle version to 6.8

soywiz (approved):
LGTM and looks reasonable the plugin part

soywiz:
136C18E2-69E4-4CD7-B3C5-A7A085126DFF

Thanks!

PR #120, opened 2021-01-15
SceneTest

serandel:
Adds sceneTest method to ViewsForTesting

### DIFF
--- a/korge/src/commonMain/kotlin/com/soywiz/korge/tests/ViewsForTesting.kt
+++ b/korge/src/commonMain/kotlin/com/soywiz/korge/tests/ViewsForTesting.kt
@@ -128,27 +128,6 @@ open class ViewsForTesting(
         )
     }
 
-	//@Suppress("UNCHECKED_CAST")
-	//fun <T : Scene> testScene(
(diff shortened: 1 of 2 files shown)


serandel:
For documentation, I'm thinking about adding the example from ViewsForTestingTest as "Example 2" at the end of the Views section in https://korlibs.soywiz.com/korge/reference/testing/. Perhaps change the test so instead of asserting over a boolean we do it over some fooView.isVisibleToUser()? That could be more clear.

soywiz:
sceneTest

Thanks!

soywiz:
> For documentation, I'm thinking about adding the example from ViewsForTestingTest as "Example 2" at the end of the Views section in https://korlibs.soywiz.com/korge/reference/testing/. Perhaps change the test so instead of asserting over a boolean we do it over some fooView.isVisibleToUser()? That could be more clear.

Sure. Feel free to add it here!: https://github.com/korlibs/korlibs.soywiz.com/blob/51d2a4308472f1f192e21d583832c79e2a7ae5e0/korge/reference/testing/index.md

PR #126, opened 2021-01-21
Do not depend on atomicfu on Klogger to prevent kotlin-reflect issues on Android

soywiz:
Continuation of https://github.com/korlibs/korge-next/pull/116


soywiz:
kotlin-reflect-atomicRef

Thanks, Carlos!

PR #125, opened 2021-01-21
Add support for binding Length units to properties: points, cm, inches, percentages or values depending on the viewport dimensions

soywiz:
@pablocaviglia-baires

PR #127, opened 2021-01-21
Added LengthDelegatedProperty to support defining length based delegated properties

soywiz:
@pablocaviglia-baires 

In a break I did this to simplify unit-dependant properties


soywiz:
Continuation on https://github.com/korlibs/korge-next/pull/125

pablocaviglia-baires:
excellent news, much more idiomatic now

PR #128, opened 2021-01-21
override pixelsPerInch on KorgeAndroidView

pablocaviglia-baires:


soywiz:
dry

Thanks, Pablo!

As you suggested, we will try to share all the code between the activity and view as possible later :)

PR #129, opened 2021-01-22
Minor optimization: Added checks to property setters in View.kt

jfbilodeau:
I'm using View.tint a fair amount in my current projects and noticed that changing the tint causes invalidateColorTransform() to be called when the property is set, even if the value is not changed. Adding a couple of if-statements sped up the code a bit at my end.


soywiz:
Thanks, @jfbilodeau !

view_alpha

PR #132, opened 2021-01-24
Fixes https://github.com/korlibs/korge/issues/330

soywiz:
There is a Robot usage on the JVM to ensure the window is focused,
but that produces an initial click event


soywiz:
misleading_click

Thanks, Carlos!

PR #135, opened 2021-01-24
Specify stencil bits on apple platforms

soywiz:
Should fix https://github.com/korlibs/korge-samples/issues/37


soywiz:
default_opengl_context_stencil_bits_0

Thanks, Carlos!

PR #143, opened 2021-01-31
Replace kotlin.math.min/max functions with custom ones

soywiz:
Details here: https://youtrack.jetbrains.com/issue/KT-40689


soywiz:
max2

Thanks, Carlos!

PR #144, opened 2021-02-01
Minor performance optimization for JS

jfbilodeau:
Greetings, I found a surprising performance bottleneck in a for loop.

Going from:
```kotlin
for (idx in 0 until min(icount, array.isize)) addIndex(vertexCount + array.indices[idx])
``` 
to
```kotlin
val minValue = if(icount < array.isize) icount else array.isize

for (idx in 0 until minValue) addIndex(vertexCount + array.indices[idx])
```
made a surprising difference for me.

Before:
image
After:
image

### DIFF
--- a/kmem/src/jsMain/kotlin/com/soywiz/kmem/BufferJs.kt
+++ b/kmem/src/jsMain/kotlin/com/soywiz/kmem/BufferJs.kt
@@ -20,48 +20,48 @@ actual typealias DataBuffer = DataView
 actual fun MemBuffer.getData(): DataBuffer = DataView(this)
 actual inline val DataBuffer.mem: MemBuffer get() = this.buffer
 actual fun DataBuffer.getByte(index: Int): Byte = this.getInt8(index)
-actual fun DataBuffer.setByte(index: Int, value: Byte): Unit = run { this.setInt8(index, value) }
+actual fun DataBuffer.setByte(index: Int, value: Byte): Unit = this.setInt8(index, value)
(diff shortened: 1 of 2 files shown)


SerVB:
I guess it should be solved like this: https://github.com/korlibs/korge-next/pull/143

jfbilodeau:
> I guess it should be solved like this: #143

Yikes! I though the min in here was an isolated incident, but I guess not. Thanks for letting me know.

soywiz:
Can you fix the merge conflict, maybe we can integrate the `run {` and `.asDynamic()` patches

soywiz:
Another huge performance hit is that ArrayList.get is calling `isObject` + `isInstanceOf`, and I don't know why. The `isInstanceOf` could be faster (I used to apply a patch in the past for the non-IR version).

You can try:

```
./gradlew :samples:bunnymark:runJs
```



With this patch:



We can get twice the bunnies with the same FPS:

jfbilodeau:
> Can you fix the merge conflict? Maybe we can integrate the `run {` and `.asDynamic()` patches

Sorry, I'm not seeing any merge conflicts at my end. Where are you seeing a conflict? Thanks!

image

soywiz:
this_branch_cannot_be_rebased_due_conflicts

jfbilodeau:
I'm going to close this pull request since I suspect FastArrayList & the new min/max will solve that silly performance issue.

Thanks!

soywiz:
I have tried and bunnymark is now comparable on JS and JVM. I hope these are temporal workarounds and they fix both issues in a future version

PR #145, opened 2021-02-02
Added FastArrayList prevenging ArrayList.get and .set from calling isObject + jsInstanceOf on the JS target

soywiz:
@jfbilodeau @SerVB 

Related to #144 

The asDynamic on JS prevents the checked cast.

Compared to:


soywiz:
Reported here: https://youtrack.jetbrains.com/issue/KT-44662

soywiz:
slow_array_list

Thanks, Carlos!

PR #146, opened 2021-02-02
Some work on #141, takes display hertz into account on Java/AWT

soywiz:
Supporting screens with higher rates than 60


soywiz:
4k_144

Thanks, Carlos!

PR #151, opened 2021-02-04
Update desktop main

jfbilodeau:
Minor fix to compile native on Linux and Mac (and probably Windows as well!)


korge-gradle-plugin/src/main/kotlin/com/soywiz/korge/gradle/targets/desktop/Desktop.kt
```diff
@@ -48,7 +48,7 @@ val Project.prepareKotlinNativeBootstrap: Task get() = tasks.createOnce("prepare
                 line("fun main(args: Array<String>): Unit = RootGameMain.runMain(args)")
                 line("object RootGameMain") {
                     line("fun runMain() = runMain(arrayOf())")
-                    line("@Suppress(\"UNUSED_PARAMETER\") fun runMain(args: Array<String>): Unit = com.soywiz.korio.Korio { ${korge.realEntryPoint}() }")
+                    line("@Suppress(\"UNUSED_PARAMETER\") fun runMain(args: Array<String>): Unit = com.soywiz.korio.Korio { ${korge.realEntryPoint}(args) }")
```
soywiz, line 51: Did this fail to you without this?

jfbilodeau: It fails for me:
```
$ ./gradlew
[...]
e: /.../shared/build/platforms/native-desktop/bootstrap.kt: (5, 103): No value passed for parameter 'args'
[...]
```

soywiz: I see. Maybe you have a main that includes an `args: Array<String>` parameter? Korge main shouldn't have that parameter.

Check the hello-world signature:

https://github.com/korlibs/korge-hello-world/blob/30ea6c36487c2f3c873cd5665fb3e5d4d1a1117b/src/commonMain/kotlin/main.kt#L11

Maybe we should state that on the FAQ since I believe this is not obvious and maybe not well documented.

Do you need the program arguments? If that's the case we could try to store them in the coroutines context to be able to retrieve them later.

The problem with changing this now, is that it would break all the existing code that doesn't have the args argument. We could support both signatures maybe by adding a flag here https://github.com/korlibs/korge-next/blob/4b0209340f40ad712513bc42f8650718c5e81ad2/korge-gradle-plugin/src/main/kotlin/com/soywiz/korge/gradle/KorgeExtension.kt#L88 specifying wheter your main has args or not.
In the future we could make an IR plugin to detect the signature automatically

jfbilodeau: Nope. Here's one of my main that does not compile without the patch:
```kotlin
import com.soywiz.korge.Korge

suspend fun main() {
    println("Starting App...")
    Korge(Korge.Config(module = AppModule))
}
```

jfbilodeau: Are you sure it would break existing code? I created a new project using the Korge template. It works without any changes with plugin version 2.0.0.1 but when I switch to 2.0.0.6, it fails to compile native. Applying the above patch makes it work. Any idea why?

jfbilodeau: ...actually, forget about it. It's working intermittently, so it's probably something on my machine. Sorry for the trouble!

soywiz: No problem, I appreciate your contribution!

I think we can improve this somehow, or at least document it. I think the main problem was to reference a main function that can be on the main package from another one, since on K/N we cannot run a `suspend fun main` and we need that to implicitly get some info from some targets like on Android that we need to propagate the android context

PR #152, opened 2021-02-04
fix error during fast repetition of touch events on android devices

pablocaviglia-baires:
also preventive code on android korviview to peacefully continue the execution when a surface lost its parent


soywiz:
Thanks Pablo!

PR #158, opened 2021-02-06
Fix for UI combox box overlap

jfbilodeau:
Fix for UIComboBox.kt where the drop-down list would be hidden under other components. Also changed the use of the .height property to .scaledHeight as suggested by the deprecation warning.


soywiz:
overlapping

Thanks, @jfbilodeau !

jfbilodeau:
> overlapping
> 
> Thanks, @jfbilodeau !

Glad I could help. Keep up the great work!

PR #159, opened 2021-02-07
Some performance tweak for js

jfbilodeau:
Massive patch with a number of minor performance improvement for KorgeJS by looking for bottlenecks.

- Replaced lots of callback with inline code
- Avoided a couple of Int->Double conversion

Many of the tweaks are focus on how I use Korge on my codebase, but I hope it can be useful for others!


korge/src/commonMain/kotlin/com/soywiz/korge/html/Html.kt
```diff
@@ -71,7 +71,7 @@ object Html {
 
 		object Identity : MetricsProvider {
 			override fun getBounds(text: String, format: Format, out: Rectangle) {
-                out.setTo(0, 0, text.length, 1)
+                out.setTo(0.0, 0.0, text.length, 1.0)
```
soywiz, line 74: This was using the integral overload. `text.length` was an integer too.  This would end using the inline `Number` overload that should be deprecated temporarily since we had performance issues with that kind of functions on Kotlin/Native.

So if you plan to change it to use the Double overload you should do:

```
out.setTo(0.0, 0.0, text.length.toDouble(), 1.0)
```

I will mark the `Rectangle.setTo` with mixed numbers deprecated for now to avoid confusions, until this issue is resolved: https://youtrack.jetbrains.com/issue/KT-39136

korma/src/commonMain/kotlin/com/soywiz/korma/geom/Rectangle.kt
```diff
@@ -183,6 +186,13 @@ data class Rectangle(
         setTo(kotlin.math.ceil(x), kotlin.math.ceil(y), kotlin.math.ceil(width), kotlin.math.ceil(height))
         return this
     }
+
+    fun zeroSelf() {
+        x = 0.0
+        y = 0.0
+        width = 0.0
+        height = 0.0
+    }
```
soywiz, line 195: There is a `clear` function already that does this. But maybe a name like `setToZero` would fit better the naming convention here. What do you think? Maybe you can update the clear function to call setToZero and maybe mark clear as deprecated.

korma/src/commonMain/kotlin/com/soywiz/korma/geom/Matrix.kt
```diff
@@ -134,21 +137,21 @@ data class Matrix(
     fun prescale(sx: Float, sy: Float = sx) = prescale(sx.toDouble(), sy.toDouble())
     fun prescale(sx: Int, sy: Int = sx) = prescale(sx.toDouble(), sy.toDouble())
 
-    fun translate(dx: Double, dy: Double) = this.apply { this.tx += dx; this.ty += dy }
+    fun translate(dx: Double, dy: Double): Matrix { this.tx += dx; this.ty += dy; return this }
     fun translate(dx: Float, dy: Float) = translate(dx.toDouble(), dy.toDouble())
     fun translate(dx: Int, dy: Int) = translate(dx.toDouble(), dy.toDouble())
 
-    fun pretranslate(dx: Double, dy: Double) = this.apply { tx += a * dx + c * dy; ty += b * dx + d * dy }
+    fun pretranslate(dx: Double, dy: Double)  { tx += a * dx + c * dy; ty += b * dx + d * dy }
```
soywiz, line 144: Missing `return this`

korma/src/commonMain/kotlin/com/soywiz/korma/geom/Matrix.kt
```diff
@@ -134,21 +137,21 @@ data class Matrix(
...
     fun pretranslate(dx: Float, dy: Float) = pretranslate(dx.toDouble(), dy.toDouble())
     fun pretranslate(dx: Int, dy: Int) = pretranslate(dx.toDouble(), dy.toDouble())
 
-    fun prerotate(angle: Angle) = this.apply {
+    fun prerotate(angle: Angle) {
```
soywiz, line 148: Missing `return this`

korma/src/commonMain/kotlin/com/soywiz/korma/geom/Matrix.kt
```diff
@@ -134,21 +137,21 @@ data class Matrix(
...
         val m = Matrix()
         m.rotate(angle)
         this.premultiply(m)
     }
 
-    fun preskew(skewX: Angle, skewY: Angle) = this.apply {
+    fun preskew(skewX: Angle, skewY: Angle) {
```
soywiz, line 154: Missing `return this`

korma/src/commonMain/kotlin/com/soywiz/korma/geom/Matrix.kt
```diff
@@ -92,7 +95,7 @@ data class Matrix(
         return this
     }
 
-    fun rotate(angle: Angle) = this.apply {
+    fun rotate(angle: Angle) {
```
soywiz, line 98: Missing `return this`

korgw/src/commonMain/kotlin/com/soywiz/korag/OpenglAG.kt
```diff
@@ -207,7 +207,7 @@ abstract class AGOpengl : AG() {
         val currentRenderBuffer = this.currentRenderBuffer ?: return
         if (currentRenderBuffer === mainRenderBuffer) {
             var realScissors: Rectangle? = finalScissorBL
-            realScissors?.setTo(0, 0, realBackWidth, realBackHeight)
+            realScissors?.setTo(0.0, 0.0, realBackWidth, realBackHeight)
```
soywiz, line 210: `realBackWidth` and `realBackHeight` are integers. Same issue described above.

korma/src/commonMain/kotlin/com/soywiz/korma/geom/Rectangle.kt
jfbilodeau: Oups! My bad. Let's keep the `clear()` function. I'll replace it with `clear()` and get rid of `zeroSelf()`

korge/src/commonMain/kotlin/com/soywiz/korge/view/Container.kt
```diff
@@ -220,15 +220,17 @@ open class Container : View(true) {
...
-                bb.add(tempRect)
-            }
-            bb.getBounds(out)
-        //}
+		bb.reset()
+//		forEachChild { child: View ->
+//			child.getBounds(this, tempRect)
+//			bb.add(tempRect)
+//		}
+        // Use regular for loop to avoid function callback in JS
```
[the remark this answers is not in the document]
jfbilodeau: You're right, the generated JS code does not (always) use callbacks, but is not always as efficient. The optimizations I'm proposing were based on the profiling I did in Chrome and Firefox based on my code, and these are some of the bottlenecks I encountered and that made a positive difference for me. What do you think?

korgw/src/commonMain/kotlin/com/soywiz/korag/OpenglAG.kt
jfbilodeau: Thanks for letting me know.

korma/src/commonMain/kotlin/com/soywiz/korma/geom/Matrix.kt
jfbilodeau: Doh! Sorry.

jfbilodeau:
If that's OK by you, I'll push an update soon.

korge/src/commonMain/kotlin/com/soywiz/korge/view/Container.kt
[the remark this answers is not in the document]
jfbilodeau: BTW, here's what I get for the same method without the `forEachChild()`

```javascript
  Container.prototype.getLocalBoundsInternal_41 = function (out) {
    this._bb_1.reset_0();
    var tmp0_iterator = this._get_children__19().iterator_73();
    while (tmp0_iterator.hasNext_54()) {
      var child = tmp0_iterator.next_61();
      child.getBounds$default_49(this, this._tempRect_5, false, false, 12, null);
      Unit_getInstance();
      this._bb_1.add_77(this._tempRect_5);
      Unit_getInstance();
    }
    this._bb_1.getBounds_3(out);
    Unit_getInstance();
  };
```
It seems a bit terser.

jfbilodeau:
I've updated the push according to your suggestions. Let me know what you think! Thanks!

korge/src/commonMain/kotlin/com/soywiz/korge/view/Container.kt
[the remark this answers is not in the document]
soywiz: As far as I know that uses an iterator, that in theory performs allocations.

Despite of that, there is a concern that if the children list is updated while getting the bounds, the iterator might fail unless you create a copy of the children beforehand.
Children might be updated for example in the `Text` view that lazily adds children if required to both render and computing its bounds. 

I would keep this function as it is for now, and keep the removal of `run { }` and `this.apply`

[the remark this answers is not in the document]
jfbilodeau: I brought back the original `forEachChild` in Container.kt. Hope that works for you!

soywiz (approved):

soywiz:
Thanks @jfbilodeau !

PR #164, opened 2021-02-09
Make smoothing of TiledMapView customizable

jobe-m:
I am currenlty porting my "retro" looking game with pixel-art graphics from
libGDX to Korge. The graphics are upscaled by factor 5 on the screen.
Now when looking into TileMapView I realized that the TileMapView is
always rendered with smoothing on. Thus my level map from Tiled looks
always blurred/smoothed.

This Pull request fixes this by making smoothing configurable. Here is
the git message text:

Rendering of a tiled map view is hardcoded to be done with smoothing
enabled. Smoothing property is not accessible because it is set in an
internal object which is not reachable from outside. This change
makes the smoothing parameter for the drawVertices call configurable
on creation of the TiledMapView object.

An unnecessary "val" string was cleaned up in the parameter list of
TiledMapView, too.

### DIFF
--- a/korge/src/commonMain/kotlin/com/soywiz/korge/view/tiles/TileMap.kt
+++ b/korge/src/commonMain/kotlin/com/soywiz/korge/view/tiles/TileMap.kt
@@ -13,23 +13,22 @@ import com.soywiz.korim.color.*
 import com.soywiz.korma.geom.*
 import kotlin.math.*
 
-inline fun Container.tileMap(map: IntArray2, tileset: TileSet, repeatX: TileMap.Repeat = TileMap.Repeat.NONE, repeatY: TileMap.Repeat = repeatX, callback: @ViewDslMarker TileMap.() -> Unit = {}) =
-	TileMap(map, tileset).repeat(repeatX, repeatY).addTo(this, callback)
+inline fun Container.tileMap(map: IntArray2, tileset: TileSet, repeatX: TileMap.Repeat = TileMap.Repeat.NONE, repeatY: TileMap.Repeat = repeatX, smoothing: Boolean = true, callback: @ViewDslMarker TileMap.() -> Unit = {}) =
+	TileMap(map, tileset, smoothing).repeat(repeatX, repeatY).addTo(this, callback)
 
-inline fun Container.tileMap(map: Bitmap32, tileset: TileSet, repeatX: TileMap.Repeat = TileMap.Repeat.NONE, repeatY: TileMap.Repeat = repeatX, callback: @ViewDslMarker TileMap.() -> Unit = {}) =
-	TileMap(map.toIntArray2(), tileset).repeat(repeatX, repeatY).addTo(this, callback)
+inline fun Container.tileMap(map: Bitmap32, tileset: TileSet, repeatX: TileMap.Repeat = TileMap.Repeat.NONE, repeatY: TileMap.Repeat = repeatX, smoothing: Boolean = true, callback: @ViewDslMarker TileMap.() -> Unit = {}) =
+	TileMap(map.toIntArray2(), tileset, smoothing).repeat(repeatX, repeatY).addTo(this, callback)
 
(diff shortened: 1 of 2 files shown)


korge/src/commonMain/kotlin/com/soywiz/korge/view/tiles/TileMap.kt
```diff
@@ -13,23 +13,22 @@ import com.soywiz.korim.color.*
...
 
 @PublishedApi
 internal fun Bitmap32.toIntArray2() = IntArray2(width, height, data.ints)
 
 @OptIn(KorgeInternal::class)
-open class TileMap(val intMap: IntArray2, val tileset: TileSet) : View() {
+open class TileMap(val intMap: IntArray2, val tileset: TileSet, val smoothing: Boolean) : View() {
     private var contentVersion = 0
-	constructor(map: Bitmap32, tileset: TileSet) : this(map.toIntArray2(), tileset)
+	constructor(map: Bitmap32, tileset: TileSet, smoothing: Boolean) : this(map.toIntArray2(), tileset, smoothing)
```
soywiz, line 28: To be backwards compatible, can you add default values to this constructors too?
The existing property was mutable, can you mark it as `var` instead of `val`?

soywiz (commented):
Nice! This is good in the case you want to have mixed content at different scales.
If you want your whole game to render at a specific resolution. You can use `filter = IdentityFilter.Nearest`:
https://github.com/korlibs/korge-next/blob/895be027a00088b31a9a6b3b72341a24332f74ac/korge/src/commonMain/kotlin/com/soywiz/korge/view/filter/IdentityFilter.kt#L14

either in the stage, or in a `FixedSizeContainer` or `Container`, that is scaled. With that filter applied, your game will be renderer into a texture at the size of the container, and then that texture will be scaled later, which is normally what you want for this kind of games

jobe-m:
> In the case it is useful for you (since I haven't added it to the documentation yet): If you want your whole game to render at a specific resolution. You can use `filter = IdentityFilter.Nearest`
> either in the stage, or in a `FixedSizeContainer` or `Container`, that is scaled. With that filter applied, your game will be renderer into a texture at the size of the container, and then that texture will be scaled later, which is normally what you want for this kind of games

Thanks for the hint. I will try it. In fact this sounds exactly what I need :)

soywiz:
Great. In addition to that, I think your contribution is valuable and makes sense. Can you add the default parameter + make the smoothing field mutable as it was, so everything keeps source-compatible? Thanks!

jobe-m:
Let me know if it is ok so

korge/src/commonMain/kotlin/com/soywiz/korge/view/tiles/TileMap.kt
```diff
@@ -23,7 +23,7 @@ inline fun Container.tileMap(map: Bitmap32, tileset: TileSet, repeatX: TileMap.R
 internal fun Bitmap32.toIntArray2() = IntArray2(width, height, data.ints)
 
 @OptIn(KorgeInternal::class)
-open class TileMap(val intMap: IntArray2, val tileset: TileSet, val smoothing: Boolean) : View() {
+open class TileMap(val intMap: IntArray2, val tileset: TileSet, var smoothing: Boolean = true) : View() {
     private var contentVersion = 0
 	constructor(map: Bitmap32, tileset: TileSet, smoothing: Boolean) : this(map.toIntArray2(), tileset, smoothing)
```
soywiz, line 28: Can you add the default parameter to this constructor too?

jobe-m: Argh, yes ;)

soywiz:
pixel2

pixelated

Awesome. Thanks @jobe-m !

jobe-m:
You're welcome! :)

PR #165, opened 2021-02-11
Bump JNA to 5.7.0

emign:
Bump JNA Version to 5.7.0 for Apple Silicon Support
Fixes: https://github.com/korlibs/korge/issues/335

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -7,7 +7,7 @@ kotlinVersion=1.4.30
 coroutinesVersion=1.4.2
 androidBuildGradleVersion=4.0.1
 proguardVersion=6.2.2
-jnaVersion=5.6.0
+jnaVersion=5.7.0
 jcodecVersion=0.2.5
 
 playServicesAdsVersion=16.0.0
@@ -19,14 +19,12 @@ android.enableJetifier=true
 
 enableKotlinNative=true
 enableKotlinNativeLinux=true
-enableKotlinMobile=true
-enableKotlinAndroid=true
+enableKotlinMobile=trueenableKotlinAndroid=true
 
 #enableKotlinNative=false
 #enableKotlinNativeLinux=false
 #enableKotlinMobile=false
 #enableKotlinAndroid=false
-
 kotlin.native.disableCompilerDaemon=true
 
 kotlin.mpp.enableGranularSourceSetsMetadata=true
--- a/gradle.properties
+++ b/gradle.properties
@@ -19,7 +19,9 @@ android.enableJetifier=true
 
 enableKotlinNative=true
 enableKotlinNativeLinux=true
-enableKotlinMobile=trueenableKotlinAndroid=true
+enableKotlinMobile=true
+enableKotlinAndroid=true
+
 
 #enableKotlinNative=false
 #enableKotlinNativeLinux=false
--- a/korge-gradle-plugin/src/main/kotlin/com/soywiz/korge/gradle/BuildVersions.kt
+++ b/korge-gradle-plugin/src/main/kotlin/com/soywiz/korge/gradle/BuildVersions.kt
@@ -13,7 +13,7 @@ object BuildVersions {
 	const val KORGW = "2.0.0.999"
 	const val KORGE = "2.0.0.999"
 	const val KOTLIN = "1.4.30"
-    const val JNA = "5.6.0"
+    const val JNA = "5.7.0"
 	const val COROUTINES = "1.4.2"
 	const val ANDROID_BUILD = "4.0.1"
 }


emign:
Only tested on my machine with Azul aarch64 1.8 JDK

soywiz:
And with this, everything works on Apple Silicon? If that's the case, that's amazing! Tough could make sense since there is only JNA and not JNI at all.

emign:
Well. It compiles a working JVM target for the aarch64 JDK/JVM from azul. 
That fixes the mentioned issue at least :)

What else there is to be discovered: I don't know :)

I wanted to introduce a github action for apple silicon for korlibs but there isn't any at the moment.

soywiz:
For the record, this is the merged PR: https://github.com/java-native-access/jna/pull/1297

BTW Nico, Im just curious, have you tried the sprites10k sample, or the bunnymark one from this repo and have some numbers of CPU usage/fps between using an emulated x64 java runtime vs using the native arm64 version. As far as I know the M1 is lightning fast, and having the GPU and embedded memory on the SoC + 5nm lithography this should be a beast

emign:
Sorry I don't have anything objectively or scientifically accurate. The debug window (F7) cannot be opened on macOS 11.2 because it freezes the JVM app.

JVM: 100,000 bunnies: around 42-44% CPU 
JVM aarch64 with 100,000 bunnies: around 40-42%

The compile times are SIGNIFICANTLY shorter on a native aarch64 JDK than the x86 one. Im speaking of factor 10x for the bunny mark.

soywiz:
Awesome. And I asume you are talking about a clean build right, were not using prebuilt classes and stuff?

Bunnymark should be both GPU and CPU-bounded I guess, and for JNA I cannot use the native approach, since some of the OpenGL symbols are dynamically loaded via another opengl call, and I don't know a way to rebind them via code, though I guess is only affected by the number of batches, and in bunnymark it is pretty small.

But weill, It is using alpha blending, and other things that could be optimized. Have you tried to not use a retina resolution and see if the FPS changes? I usually use this: https://github.com/avibrazil/RDM to switch between retina and non-retina resolutions, to check that everything works. Retina has 4 times the number of pixels, and in latests versions it also does a 4x MSAA I believe (not sure if that happens on MacOS too, but I guess we could drop it when using retina resolutions)

emign:
I could do some more research when it is implemented in KorGE itself. That would make testing much easier for me. IntelliJ in the aarch64 version does not like the HUGE KorGE-next Project. Maybe its Gradle too. It will have aarch64 in Gradle 7.0

soywiz:
apple_m1

apple_m1_2

Seems to work properly on my machines, thanks Nico! :)

emign:
Azul apple silicon JDK: FPS drop below 50 at around 250k bunnies. under 2 seconds on third built till window
Adoptionen Rosetta/x64: FPS drop below 50 at around 90k bunnies. 8 seconds on third built till window

soywiz:
Thats around 3-4x, nice. There are some improvements pending in both gpu and cpu

soywiz:
IDEA is suffering a lot with korge-next. I have simplified it a bit by not including the intellij plugin here.

Regarding to optimizations:

It would be possible to use geometry shaders, emitting just points and generating quads from it, effectively reducing the GPU bandwidth requirements to almost 1/4, and not performing lots of floating on the CPU.
But KorGE is not going to support geometry shaders in the mid term because WebGL1 or even 2 doesnt support it, so until something like WebGPU similar to vulkan/metal is available and mainstream I wont probably add support for something more advanced.

In the meantime there is a trick to achieve a similar effect that theoretically requires twi extensions widely available:

https://developer.mozilla.org/en-US/docs/Web/API/ANGLE_instanced_arrays

https://developer.mozilla.org/en-US/docs/Web/API/OES_texture_float

But would require a value greater than one for this property: MAX_VERTEX_TEXTURE_IMAGE_UNITS 

We could create a quad, and render lots of instances on it, then use a texture to read all the properties for all the stuff, but requires texture sampling on the vertex shader that is not always available.

We can also try to reorder how memory of the objects is allocated and stored to be as much contiguous as possible

soywiz:
@emign Can you try master again on your Apple M1? You should be able to place a few more bunnies :)

emign:
This brings some funny problems with recurring frame drops. See these videos:

https://youtu.be/5zZLRCo6vu4

https://youtu.be/TOC3Ez4Wm3k

soywiz:
Videos are private

emign:
fixed

soywiz:
Have you master updated?

emign:
Its a completely new git clone

soywiz:
I had that issue before and fixed it with this commit:
https://github.com/korlibs/korge-next/commit/087021edbf916109777e68aff6c0cefad837719a

That's why I asked. So if you perform a `git pull` it doesn't bring new commits?

emign:
Head was ob 087021e
I pulled 08314d a6ebf72 now. Retesting

emign:
https://youtu.be/MQkFEppIBd0

soywiz:
Just tried on an intel mac, and seems to work properly. If you reduce to `, batchMaxQuads = 4000` the problem persists, and sprites stop displaying at 4000?

emign:
Better:
https://youtu.be/z70qZovsAww

soywiz:
Still that shows artifacts, it is a bit strange. Can you try to run it on JS? In this effort I tried to optimize all the targets, and each one has its own peculiarities. But would want to see if the JS target displays those artifacts on your computer

emign:
JS does not show artifacts. But I believe the z-index of the pouring bunnies does not look right. The stream of new bunnies plops behind the rest just to reappear.

https://youtu.be/ygGLuXYWg40

Thats on `BatchBuilder2D.MAX_BATCH_QUADS)`

soywiz:
Have you `git pull` I think I have fixed the issue you had here: https://github.com/korlibs/korge-next/commit/ee1b313d0e747a82ce10cfda64c31a0a8b17f0c6

emign:
just saw that after my answer. building as we speak

emign:
yes its fixed
wohoo

120k bunnies aarch64 JDK to drop under 50 FPS. But that's on the MacBook Air . 
Have to test it on the mini later/tomorrow (the results from above are from the mini)

soywiz:
That's much worse then? Or is that rosetta?

> Azul apple silicon JDK: FPS drop below 50 at around 250k bunnies. under 2 seconds on third built till window
> Adoptionen Rosetta/x64: FPS drop below 50 at around 90k bunnies. 8 seconds on third built till window

emign:
yes it is worse for the aarch64 jdk.
Good message is, that the x86/rosetta version is at 120k too now.
But the aarch64 version regressed

soywiz:
If VisualVM works on aarch64, could you profile the old version and the new one when you have time and send me the snapshot `.nps` files for 20 seconds of CPU sampling with 400K sprites on aarch64 old faster version and new slower version?



> Pressing the `Sampler` -> `CPU`, then `Snapshot`, then the `diskette` `Export Snapshot data`  to save it

emign:
ofc. But I cannot do it today.
Well tomorrow is in 19 minutes but I mean this night :)

emign:
I tested with 7d92c54 on the mini again.
If you didn't make any changes 

**Drop under 50 FPS**
Mac mini x86/rosetta JDK: around 200k Sprites
Mac mini aarch64 JDK: around 330k Sprites
MacBook Air x86/rosetta JDK: around 200k Sprites
MacBook Air aarch64 JDK: around 330k Sprites
(I tried it with power attached and on battery)

So the results are consistent again, which makes sense, because they habe the exact same CPU/GPU. Did you do any optimizations in between?

soywiz:
Hey @emign ! How are you doing?

Could you by chance try: `./gradlew :samples:bunnymark-fast:runJvm` on the Apple M1? And try to resize the window to make it smaller. The -fast sample should now be GPU-bounded and I wonder if it can reach 60fps with 800K bunnies

PR #172, opened 2021-02-22
Add support in sprite animations for using a trimmed atlas

jobe-m:
When saving a sprite atlas in "trimmed" mode e.g. with the (awesome)
aseprite than it should be possible to use this "space optimized" atlas
in Korge when setting up a sprite animation object. Trimmed sprite atlases
can save a lot empty unused pixel space around single sprite animation
frames.

For that the drawing function needs the info to shift the sprite frames
within their original size. This is done by modifying the anchorX and
anchorY values of the sprite. So when playing a sprite animation the
anchors needs to be updated together with the bitmap slice from the
texture.

For calculating a list of anchorXYs for a sprite animation the region
information of the unterlying sprite altas is needed in the Sprite
class. This comes into the class with the regionDetails list property.
That list contains for each animation frame the original size before
trimming (sourceSize) as well as the offset and the trimmed smaller
real size (spriteSourceSize). With that details the needed offset is
calculated and saved in above mentioned list (currentAnchorXY).

The original behaviour of anchorX and anchorY in the sprite is kept.
That means when setting the anchors of a sprite object by the Anchorable
API they work as before. That is acomplished by overriding them in the
Sprite class. So when setting the anchor of a Sprite which uses a
trimmed atlas all underlying values in the anchorXY list are adjusted as
well.

It was checked that the following API calls are working as before also
for trimmed sprite animations:

- sprite(spriteAnimation, anchorX = x1, anchorY = y1)
- sprite(spriteAnimation).anchor(x1, y1)
- sprite(spriteAnimation).apply {
    anchorX = x1
    anchorY = y1
  }

When using SpriteAnimation.invoke() function which specifies an
animation out of a texture by rows and columns then the regionDetails
property is set to an empty list. Region info is not needed in that
case. The empty list does not add much additional overhead on drawing.
In this case the anchorX and anchorY values are not taken from the
currentAnchorXY list in the Sprite class as that list will be empty.

Some small additional reformatting of code was done (auto-formating by
the editor).

### DIFF
--- a/korge/src/commonMain/kotlin/com/soywiz/korge/view/RectBase.kt
+++ b/korge/src/commonMain/kotlin/com/soywiz/korge/view/RectBase.kt
@@ -44,9 +44,9 @@ open class RectBase(
 	protected open val bheight get() = 0.0
 
     //@KorgeInternal
-    override val anchorDispX get() = (anchorX * bwidth)
+    override val anchorDispX get() = when { baseBitmap.trimmed -> (anchorX * baseBitmap.virtFrame.width - baseBitmap.virtFrame.x) else -> (anchorX * bwidth) }
     //@KorgeInternal
-    override val anchorDispY get() = (anchorY * bheight)
+    override val anchorDispY get() = when { baseBitmap.trimmed -> (anchorY * baseBitmap.virtFrame.height - baseBitmap.virtFrame.y) else -> (anchorY * bheight) }
 
     protected open val sLeft get() = -anchorDispX
 	protected open val sTop get() = -anchorDispY
--- a/korge/src/commonMain/kotlin/com/soywiz/korge/view/RectBase.kt
+++ b/korge/src/commonMain/kotlin/com/soywiz/korge/view/RectBase.kt
@@ -44,9 +44,9 @@ open class RectBase(
 	protected open val bheight get() = 0.0
 
     //@KorgeInternal
-    override val anchorDispX get() = when { baseBitmap.trimmed -> (anchorX * baseBitmap.virtFrame.width - baseBitmap.virtFrame.x) else -> (anchorX * bwidth) }
+    override val anchorDispX get() = (anchorX * bwidth)
     //@KorgeInternal
-    override val anchorDispY get() = when { baseBitmap.trimmed -> (anchorY * baseBitmap.virtFrame.height - baseBitmap.virtFrame.y) else -> (anchorY * bheight) }
+    override val anchorDispY get() = (anchorY * bheight)
 
     protected open val sLeft get() = -anchorDispX
 	protected open val sTop get() = -anchorDispY
--- a/korge/src/commonMain/kotlin/com/soywiz/korge/view/Sprite.kt
+++ b/korge/src/commonMain/kotlin/com/soywiz/korge/view/Sprite.kt
@@ -1,12 +1,9 @@
 package com.soywiz.korge.view
 
-import com.soywiz.kds.FastArrayList
 import com.soywiz.klock.*
 import com.soywiz.kmem.umod
-import com.soywiz.korim.atlas.AtlasInfo
 import com.soywiz.korim.bitmap.*
 import com.soywiz.korio.async.Signal
-import com.soywiz.korma.geom.Anchor
 import com.soywiz.korma.geom.vector.VectorPath
 
 inline fun Container.sprite(
@@ -47,7 +44,7 @@ open class Sprite(
     smoothing: Boolean = true
 ) : BaseImage(bitmap, anchorX, anchorY, hitShape, smoothing) {
     constructor(
-        bitmap: Bitmap,
+        bitmap : Bitmap,
         anchorX: Double = 0.0,
         anchorY: Double = anchorX,
         hitShape: VectorPath? = null,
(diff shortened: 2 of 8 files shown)


korge/src/commonMain/kotlin/com/soywiz/korge/view/Sprite.kt
```diff
@@ -119,11 +178,18 @@ open class Sprite(
         }
 
     private var currentAnimation: SpriteAnimation? = null
+    private var currentAnchorXY = FastArrayList<Anchor>()
```
soywiz, line 181: Since this is private and in a class that could be instantiated and accessed many times. What would you think about using `PointArrayList` instead? 
This one: https://github.com/korlibs/korma/blob/176dd4e62522d3f654ee21c383c8ecdf116b9074/korma/src/commonMain/kotlin/com/soywiz/korma/geom/PointArrayList.kt#L31

The point of that class is that you can store two X and Y values without allocating, and keeping the memory contiguous to take advantage of CPU caches.
By using Anchor you are allocating the array then separate objects for each value. Additionally, that forces Anchor to make its fields mutable

korma/src/commonMain/kotlin/com/soywiz/korma/geom/Anchor.kt
```diff
@@ -3,7 +3,7 @@ package com.soywiz.korma.geom
 import com.soywiz.korma.interpolation.Interpolable
 import com.soywiz.korma.interpolation.interpolate
 
-data class Anchor(val sx: Double, val sy: Double) : Interpolable<Anchor> {
+data class Anchor(var sx: Double, var sy: Double) : Interpolable<Anchor> {
```
soywiz, line 6: Anchor has immutable fields because it has a companion object with default values. This have several consequences:
* Might cause issues on Kotlin/Native, by having global mutable state
* Some might accidentally mutate one of the predefined instances, causing hard to catch bugs

Making Anchor mutable, would require changing this. In the case of the mutable `Matrix` class, instead of providing a global identity matrix, I'm adding a method to set a specific matrix to an identity one.

In this case, changing this would break existing code so we should keep this specific class immutable

soywiz (commented):
Awesome, thanks for the contribution. I believe we can and should extrapolate this not only to Sprites that are attached to animations, but to any image, probably `BaseImage`. What would you think about going to the `BitmapSlice` class and adding it support for padding defaulting to 0? That way we can reuse it everywhere. This is the `BitmapSlice`:
https://github.com/korlibs/korge-next/blob/9eaadf19562ce4e65a52c30713ff2f38d8a1be56/korim/src/commonMain/kotlin/com/soywiz/korim/bitmap/BitmapSlice.kt#L8
It has already a property called rotate there with the intention of supporting rotated sprites that are generated by atlas packers that further optimize space by rotating textures. Maybe we could add either a padding, or a frame rectangle, or something. Then update `BaseImage` to apply that offset when computing actual vertices.

Would that make sense to you?

jobe-m:
Yes, I guess that would make sence. I need to look into BitmapSlice and BaseImage and will come back either with a proposal or further questions or both ;)

korge/src/commonMain/kotlin/com/soywiz/korge/view/Sprite.kt
jobe-m: Thanks for pointing me towards it. I will check how to use it here.

korma/src/commonMain/kotlin/com/soywiz/korma/geom/Anchor.kt
jobe-m: All right! I will change it back to immutable. In fact I had some doupts that this change might break something else :)

jobe-m:
I think this deeper implementation looks now must cleaner as before :)
I checked again my "use cases" above and they work as well as before.

korge/src/commonMain/kotlin/com/soywiz/korge/view/RectBase.kt
```diff
@@ -44,9 +44,9 @@ open class RectBase(
 	protected open val bheight get() = 0.0
 
     //@KorgeInternal
-    override val anchorDispX get() = (anchorX * bwidth)
+    override val anchorDispX get() = when { baseBitmap.trimmed -> (anchorX * baseBitmap.virtFrame.width - baseBitmap.virtFrame.x) else -> (anchorX * bwidth) }
```
soywiz, line 47: Since there are only two branches here, could we use an if instead to be consistent with the codebase?

korim/src/commonMain/kotlin/com/soywiz/korim/bitmap/BitmapSlice.kt
```diff
@@ -5,11 +5,19 @@ import com.soywiz.kmem.*
...
+ *                     the bounds rectangle. Width and height defines the size of the virtual frame.
+ */
 abstract class BmpSlice(
     val bmpBase: Bitmap,
     val bounds: RectangleInt,
     val name: String? = null,
-    val rotated: Boolean = false
+    val rotated: Boolean = false,
+    val trimmed: Boolean = false,
+    val virtFrame: RectangleInt = RectangleInt(0, 0, bounds.width, bounds.height)
```
soywiz, line 20: Maybe virtFrame could be nullable, then trimmed be something like:
```
val trimmed get() = virtFrame != null
```

Could we also precompute the virtual width/height in both cases (trimmed and not trimmed) and store it in this class to improve CPU-cache locality / avoid one indirection when using this info?

korge/src/commonMain/kotlin/com/soywiz/korge/view/RectBase.kt
soywiz: bwidth is not being used in the trimmed version branch, and seems to be `protected open val`. Don't remember / didn't check yet the implementors but we should be sure this is working properly in several cases

soywiz (commented):
Much better and simpler! I like this approach much more :) This is something that was pending to support, so it is super great you did it. I put a few inline comments we can discuss before merging, but I'm pretty sure they can be solved easily.

korge/src/commonMain/kotlin/com/soywiz/korge/view/RectBase.kt
jobe-m: Good idea. Could think of setting up AnchorDispX and Y from local properties or members. So we wouldn't need also an if clause. But instead set the properties depending if trimmed or not. I will look into it. I am currently on the road with the kids ;)

jobe-m:
I guess two things should now be better:
- Usage of bwidth and bheight for calculating anchorDispX/Y is as before in RectBase. I thinkg it makes more sence to override in BaseImage since there a trimmed BitmapSlice is actually used. Right?
- frame offset and size is stored separately in BitmapSlice. Does this increase the CPU-cache locality now better?

soywiz:
Looks great. Thanks @jobe-m , really appreciated.

Since this is korge-next, we can cover corner cases later. At least I guess it should work with existing atlases for now, and we can try new cases later.

too_much_space

trimming_empty_space

jobe-m:
I will keep my eyes open when now starting to use Atlases. And later I will look deeper into Tiled maps and if  everything is there :)
Thanks for Korge. I am glad to help you making it better.

Issue #171, opened 2021-02-17
Sound clips get cut of on Mac OS and Windows


jfbilodeau:
Some sound clips that are >~1 second get cut off when playing native on Windows and Mac.

Please use my game as a sample. When entering the mountain the clip `./editor/src/commonMain/resources/sounds/teleport.wav` plays or when getting the 'lens,' the clip `./editor/src/commonMain/resources/sounds/teleport.wav` plays but they are cut short on native.

Thanks for you time!

jfbilodeau:
I don't know if that can help, but I found that on MacOS Native, sound works fine for a bit, then CPU usage shoots up and FPS drops. Playing without sound keeps a constant FPS. Seems to work fine on Linux.

soywiz:
That helps indeed! I believe OpenAL might have two implementations, one for sounds, and other for audio streams. While on MacOS and Windows, there is only an implementation for audio streams, that is used as fallback for other sounds to simplify the implementation. Maybe the audio stream is not being finalized when the sound is completed? Going to investigate. Thanks for the report!

soywiz:
Closing this for now. Please, let me know if you still have issues on MacOS and reopen this if required

PR #175, opened 2021-02-24
MingwX64 Block-free audio in a separate worker

soywiz:
Partial fix for https://github.com/korlibs/korge-next/issues/171


soywiz:
waveOutWrite

Thanks Carlos!

PR #179, opened 2021-02-27
[Fix] Updated BitSet to use all 32 bits of an Int

andrei-cosmin:


soywiz:
Thanks @andrei-cosmin !

nibble

4bits

soywiz:
In this codebase nobody is using data directly and hope nobody else is using it, so I guess it is ok to make it private.
Maybe we could provide a copyTo and clone methods that use this array to perform a fast copy that doesn't require copying bit by bit

PR #181, opened 2021-03-03
Supports providing a playback coroutineContext/scope when playing sounds

soywiz:
Fixes https://github.com/korlibs/korge/issues/342


soywiz:
Thanks @soywiz !

s2
s3
s1
s4

PR #182, opened 2021-03-03
Add FlagFilter

MikeSmithEU:
A FlagFilter that distorts the texture using increasing waves to the right, keeping the left-most vertical static, as if connected to a flag pole.

Sample project:
`./gradlew :samples:flag:runJvm`

Example of this filter in use:

https://user-images.githubusercontent.com/2065591/109848297-60e85980-7c50-11eb-9905-704e6a9fbaac.mov

*I have read CONTRIBUTING.md and accept contributing this to the korge project under the Apache 2.0 license*


korge/src/commonMain/kotlin/com/soywiz/korge/view/filter/FlagFilter.kt
```diff
@@ -0,0 +1,75 @@
...
+) : ShaderFilter() {
+    companion object {
+        val u_amplitude = Uniform("amplitude", VarType.Float1)
+        val u_crestCount = Uniform("crestCount", VarType.Float1)
+        val u_cyclesPerSecond = Uniform("cyclesPerSecond", VarType.Float1)
+        val u_Time = Uniform("time", VarType.Float1)
+
+        private val FRAGMENT_SHADER = FragmentShader {
+            apply {
+                val x01 = fragmentCoords01.x - ceil(abs(u_amplitude)) / (u_TextureSize.x + 2.lit * ceil(abs(u_amplitude)))
```
soywiz, line 40: Please, change `2.lit` with `2f.lit`.

Rationale:
Unfortunately it seems that some opengl drivers support automatic type casting, others not, and this is not checking yet at the DSL level (I want to make the DSL type-safe and perform verifications after building)

SerVB, line 40:
```suggestion
                val x01 = fragmentCoords01.x - ceil(abs(u_amplitude)) / (u_TextureSize.x + 2f
.lit * ceil(abs(u_amplitude)))
```

@soywiz, by the way, you can propose changes like here.

image

MikeSmithEU: Done

soywiz: Ohh 💯, awesome! Thanks didn't know that :)

soywiz:
LGTM. Thanks @MikeSmithEU !

korge-flag

PR #183, opened 2021-03-03
Bumps Kotlin to 1.5.0-M1

soywiz:
PR/branch for preparing for 1.5.0

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -2,7 +2,8 @@
 version=2.0.0.999
 #version=2.0.0
 
-kotlinVersion=1.4.31
+kotlinVersion=1.5.0-M1
+#kotlinVersion=1.4.31
 #kotlinVersion=1.4.30-RC
 coroutinesVersion=1.4.2
 androidBuildGradleVersion=4.0.1
--- a/korge-gradle-plugin/src/main/kotlin/com/soywiz/korge/gradle/BuildVersions.kt
+++ b/korge-gradle-plugin/src/main/kotlin/com/soywiz/korge/gradle/BuildVersions.kt
@@ -2,7 +2,7 @@
 package com.soywiz.korge.gradle
 
 object BuildVersions {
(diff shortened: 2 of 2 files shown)


soywiz:
Running hello world on JVM causes an issue I have to check (maybe related to IR being the default backend for JVM now):



JS doesn't work either. `http://localhost:8080/main.js` doesn't exist. So maybe the gradle plugin should be updated to get the proper JS name or something

SerVB:
Maybe M2 will fix the errors: https://blog.jetbrains.com/kotlin/2021/03/kotlin-1-5-0-m2-released-ensure-smooth-migration-to-kotlin-1-5-0

soywiz:
Let me try

soywiz:
JVM seems to work. JS doesn't display anything. Gradle plugin might have changed and will probably require some adjustments to determine the output file.

Do you know if values classes for common will be included in the final release? I have tried them but it says that it requires `@JvmInline` and can only have one single parameter. Now that all the targets are using IR by default, maybe I could try to write an IR plugin for automatic dependency injection.
For me the most exciting feature was the value classes with several properties which could change the way KorGE is used and would justify bumping to 3.0 earlier

SerVB:
> value classes with several properties

AFAIR several properties aren't available in 1.5. I wait for them too but don't know when they come.

> Do you know if values classes for common will be included in the final release?

It seems inline classes are just renamed to value for now (and a couple of improvements are done, more info [here](https://kotlinlang.org/docs/whatsnew1430.html#improved-inline-classes)). So they should be available. I'm not sure why they require JvmInline. Maybe KorGE can continue to use inline instead of value, the only drawback is that it will become marked as deprecated and be removed:

> The inline and value modifiers are now equivalent to each other. In future Kotlin versions, we’re planning to deprecate the inline modifier.

soywiz:
I have migrated to 1.5.0-RC. Found a few bugs in the process, but managed to circumvent them

PR #184, opened 2021-03-03
Make flag sample interactive

MikeSmithEU:
Small change to easily be able to change flag waving parameters in sample...


soywiz:
Ty!

PR #187, opened 2021-03-04
Add a tween sample, semi-equivalent to the animations sample

MikeSmithEU:


soywiz:
Cool. Thanks! :)

same_sample

PR #188, opened 2021-03-05
Make rectangle contains function more generic

SerVB:
This PR:
1. Allows usage of `.contains` for `IRectangle` (before it worked only for `Rectangle`).
1. Removes `.contains(that: Point)` because there is `contains(that: IPoint)` already.
1. Adds `.contains(that: PointInt)`.
1. Adds the same functions for `IRectangleInt`.
1. Adds tests for these methods.

### DIFF
--- a/korma/src/commonMain/kotlin/com/soywiz/korma/geom/Rectangle.kt
+++ b/korma/src/commonMain/kotlin/com/soywiz/korma/geom/Rectangle.kt
@@ -31,6 +31,12 @@ val IRectangle.topRight get() = Point(right, top)
 val IRectangle.bottomLeft get() = Point(left, bottom)
 val IRectangle.bottomRight get() = Point(right, bottom)
 
+operator fun IRectangle.contains(that: IPoint) = contains(that.x, that.y)
+operator fun IRectangle.contains(that: IPointInt) = contains(that.x, that.y)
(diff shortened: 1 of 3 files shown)


korma/src/commonMain/kotlin/com/soywiz/korma/geom/Rectangle.kt
```diff
@@ -302,6 +303,11 @@ fun RectangleInt.setBoundsTo(left: Int, top: Int, right: Int, bottom: Int) = set
 ////////////////////
 
 operator fun IRectangleInt.contains(v: SizeInt): Boolean = (v.width <= width) && (v.height <= height)
+operator fun IRectangleInt.contains(that: IPoint) = contains(that.x, that.y)
+operator fun IRectangleInt.contains(that: IPointInt) = contains(that.x, that.y)
+fun IRectangleInt.contains(x: Double, y: Double) = (x >= left && x < right) && (y >= top && y < bottom)
+fun IRectangleInt.contains(x: Float, y: Float) = contains(x.toDouble(), y.toDouble())
+fun IRectangleInt.contains(x: Int, y: Int) = contains(x.toDouble(), y.toDouble())
```
soywiz, line 310: Dear Internet journeyman. If you are reading these lines, you can still understand the motivation of this, but still suffer as much as I do, and your retinas are shedding from your eyes in pain, and there are no onions near. Join my cause: https://youtrack.jetbrains.com/issue/KT-39136

PR #189, opened 2021-03-05
Add missing constructors for int metrics

SerVB:
This PR adds methods for constructing metrics that I miss in my project:
1. I haven't found a way of copying immutable metrics to a mutable object, so I add three methods to SizeInt, PointInt, and RectangleInt.
1. There was no way of constructing ISizeInt, so I added it.


soywiz:
Thanks @SerVB !

operator_fun_invoke

PR #190, opened 2021-03-05
Make setTo parameter more generic

SerVB:
This PR makes it possible to use these methods with immutable arguments, as it's already possible for PointInt.


soywiz:
Thanks @SerVB !

rectangle_int

PR #191, opened 2021-03-06
Translate ordinals to Russian (korlibs/klock#153)

SerVB:
This partly addresses korlibs/korge#669.

@XJIEBYUJEK, @RezMike, could you please take a look? Does it look OK to you in terms of both Russian and implementation?

### DIFF
--- a/klock/src/commonMain/kotlin/com/soywiz/klock/KlockLocalContext.kt
+++ b/klock/src/commonMain/kotlin/com/soywiz/klock/KlockLocalContext.kt
@@ -1,0 +1,16 @@
+package com.soywiz.klock
+
+data class KlockLocalContext(val gender: KlockLocaleGender) {
+
+    companion object {
+
+        val empty = KlockLocalContext(KlockLocaleGender.Neuter)
+    }
+}
+
+fun KlockLocalContext.withGender(gender: KlockLocaleGender): KlockLocalContext = this.copy(gender = gender)
+
+enum class KlockLocaleGender {
+    Neuter,
+    Masculine,
+}
--- a/klock/src/commonMain/kotlin/com/soywiz/klock/KlockLocalContext.kt
+++ b/klock/src/commonMain/kotlin/com/soywiz/klock/KlockLocalContext.kt
@@ -1,8 +1,0 @@
-package com.soywiz.klock
-
-data class KlockLocalContext(val gender: KlockLocaleGender = KlockLocaleGender.Neuter)
-
-enum class KlockLocaleGender {
-    Neuter,
-    Masculine,
-}
--- a/klock/src/commonMain/kotlin/com/soywiz/klock/KlockLocalContext.kt
+++ b/klock/src/commonMain/kotlin/com/soywiz/klock/KlockLocalContext.kt
@@ -1,14 +1,6 @@
 package com.soywiz.klock
 
-data class KlockLocalContext(val gender: KlockLocaleGender) {
-
-    companion object {
-
-        val empty = KlockLocalContext(KlockLocaleGender.Neuter)
-    }
-}
-
-fun KlockLocalContext.withGender(gender: KlockLocaleGender): KlockLocalContext = this.copy(gender = gender)
+data class KlockLocalContext(val gender: KlockLocaleGender = KlockLocaleGender.Neuter)
 
 enum class KlockLocaleGender {
     Neuter,
--- a/klock/src/commonMain/kotlin/com/soywiz/klock/KlockLocale.kt
+++ b/klock/src/commonMain/kotlin/com/soywiz/klock/KlockLocale.kt
@@ -29,7 +29,7 @@ abstract class KlockLocale {
         }
     }
 
-    open fun getOrdinalByDay(day: Int, context: KlockLocaleContext = KlockLocaleContext()): String = ordinals[day]
+    open fun getOrdinalByDay(day: Int, context: KlockLocaleContext = KlockLocaleContext.Default): String = ordinals[day]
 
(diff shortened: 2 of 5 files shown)


XJIEBYUJEK:
Seems OK

klock/src/commonMain/kotlin/com/soywiz/klock/locale/ru.kt
```diff
@@ -1,13 +1,23 @@
...
+import com.soywiz.klock.KlockLocaleGender
 
 val KlockLocale.Companion.russian get() = RussianKlockLocale
 
 open class RussianKlockLocale : KlockLocale() {
 	companion object : RussianKlockLocale()
 
+    override fun getOrdinalByDay(day: Int, context: KlockLocalContext): String = when (context.gender) {
+        KlockLocaleGender.Masculine -> "$day-й"
+        // if feminine is ever added to KlockLocaleGender, don't forget to add implementation here: "$day-я"
```
soywiz, line 15: Let me be sure I understand: 

Are there there three ways for doing this?

```
Masculine: $day-й
Femenine: $day-я
And neutral: $day-е
```

Is that the case? And if it is, why aren't you adding femenine here too? Is it because this form won't happen on 1-31 days?

klock/src/commonMain/kotlin/com/soywiz/klock/KlockLocalContext.kt
```diff
@@ -0,0 +1,16 @@
+package com.soywiz.klock
+
+data class KlockLocalContext(val gender: KlockLocaleGender) {
```
soywiz, line 3: Maybe this could be somehow private or internal, so people can construct this directly? Since I guess adding new parameters to a constructor is not ABI-compatible, but the withGender should keep it ABI-compatible if at some point other language requires other extra context.

We could also asume no ABI compatibility here and only source-code compatibility, and either force named arguments, or just add default values, and add new parameters always at the end.
In that case, maybe we don't need the .withGender, and we can just use the `data class` `KlockLocaleContext(...).copy(gender = ...)`. What do you think?

klock/src/commonMain/kotlin/com/soywiz/klock/locale/ru.kt
SerVB: Exactly. I just couldn't find an example where this ordinal is used in femenine for "date numbers" in Russian. So I think it's redundant to add femenine to enum for now.

If femenine is added by another language or someone finds such an example in Russian, I think it'll be a right rime to add it.

klock/src/commonMain/kotlin/com/soywiz/klock/KlockLocalContext.kt
SerVB: Good point.

I think if we want to deal with ABI-compatibility, `KlockLocaleContext(...).copy(gender = ...)` can become incompatible too... Do you think it's true? I've just googled briefly and haven't found info about compatibility of data classes. If yes, then I guess we should restrict `copy` from external usage too, for example, by making this class not-data.

If you mean only source-compatibility, then copy shouldn't be the problem.

Anyways, I agree that we should restrict visibility of the constructor. I'll do it.

SerVB: @soywiz, so which variant is better?

1.

```kt
data class KlockLocalContext internal constructor(val gender: KlockLocaleGender) {

    companion object {

        val empty = KlockLocalContext(KlockLocaleGender.Neuter)
    }
}
```

2.

```
data class KlockLocalContext(val gender: KlockLocaleGender = KlockLocaleGender.Neuter)
```

The last one, right?

soywiz: I would go with 2 because of simplicity. And it can be source-level compatible, so 👍

klock/src/commonMain/kotlin/com/soywiz/klock/KlockLocale.kt
```diff
@@ -30,7 +30,7 @@ abstract class KlockLocale {
         }
     }
 
-    open fun getOrdinalByDay(day: Int, context: KlockLocalContext = KlockLocalContext.empty): String = ordinals[day]
+    open fun getOrdinalByDay(day: Int, context: KlockLocalContext = KlockLocalContext()): String = ordinals[day]
```
soywiz, line 33: One minor thing here. Could you create a companion object with something like:

val Default = KlockLocalContext()

That way in the default case it won't allocate on each call

SerVB: Sure, will do.

Also, what do you think about the name `KlockLocalContext`? [Here](https://github.com/korlibs/klock/issues/153#issuecomment-791742614) you've proposed `local` but maybe you've meant `locale` (`KlockLocaleContext`)?

If you decide it's worth renaming, I think it's better to rename it before merging, so I will do it too.

soywiz: Looks either like a typo, or my current self thinks that my other day self was drunk :)

Right now I reviewed it in a phone and I didn't notice the change lol 🤣

My criteria is that if it could be an inner class, it would be named KlockLocale.Context, without the dot KlockLocaleContext.

Local doesn't make sense to me, at least not now 🤷🏻

SerVB: Done

soywiz:
Awesome, thanks @SerVB !

20-e

PR #192, opened 2021-03-06
Get rid of underscores in IRectangle (resolve korlibs/korma#45)

### LINKED ISSUE #45: Get rid of underscores in IRectangle
Current state seems to have redundant properties:

```kotlin
interface IRectangle {
    val _x: Double
    val _y: Double
    val _width: Double
    val _height: Double
}
val IRectangle.x get() = _x
val IRectangle.y get() = _y
val IRectangle.width get() = _width
val IRectangle.height get() = _height
```

I think it can become like this:
```kotlin
interface IRectangle {
    val x: Double
    val y: Double
    val width: Double
    val height: Double
}
```

As [discussed](https://discord.com/channels/728582275884908604/728616376910217258/794909014184951828), it can be handy but low-performant on native.

SerVB:
This resolves korlibs/korma#45.

### DIFF
--- a/korma/src/commonMain/kotlin/com/soywiz/korma/geom/Rectangle.kt
+++ b/korma/src/commonMain/kotlin/com/soywiz/korma/geom/Rectangle.kt
@@ -16,6 +16,15 @@ interface IRectangle {
     }
 }
 
+@Deprecated("Properties with underscores are deprecated and will be removed soon", ReplaceWith("x"), DeprecationLevel.ERROR)
+val IRectangle._x get() = x
+@Deprecated("Properties with underscores are deprecated and will be removed soon", ReplaceWith("y"), DeprecationLevel.ERROR)
+val IRectangle._y get() = y
+@Deprecated("Properties with underscores are deprecated and will be removed soon", ReplaceWith("width"), DeprecationLevel.ERROR)
+val IRectangle._width get() = width
+@Deprecated("Properties with underscores are deprecated and will be removed soon", ReplaceWith("height"), DeprecationLevel.ERROR)
+val IRectangle._height get() = height
+
 val IRectangle.left get() = x
 val IRectangle.top get() = y
 val IRectangle.right get() = x + width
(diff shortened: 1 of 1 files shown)


korma/src/commonMain/kotlin/com/soywiz/korma/geom/Rectangle.kt
```diff
@@ -4,10 +4,10 @@ import com.soywiz.korma.internal.*
 import com.soywiz.korma.interpolation.*
 
 interface IRectangle {
```
soywiz, line 6: Maybe you can extend IPoint now?

soywiz (commented):
Nice. I think if this is not optimal right now because devirtualization is not working, this should be handled at Kotlin/Native level if possible.

Could you add `IRectangle._x`, etc. with `@Deprecated+replace with`. Not sure if someone is using it. Hopefully only used internally, but let's give it a couple of versions before removing them completely just in case

korma/src/commonMain/kotlin/com/soywiz/korma/geom/Rectangle.kt
SerVB: Seems you are talking about korlibs/korma#44. Yep, I'm planning to do it in another PR, after we deal with this one.

soywiz:
What did you think about providing deprecated extensions (maybe hidden) `_x`, `_y`, `_width` and `_height` in the end?

SerVB:
Sorry, I don't have much time now. I will do it of course, I'll add a commit to this PR, hopefully on this weekend, and ping you.

Actually, I've reacted with 👀 to note that I've seen your message and will proceed a bit later 😉

soywiz:
Sure, no rush!

SerVB:
@soywiz, I've used ERROR deprecation level; I think HIDDEN makes message and ReplaceWith useless, so I've filed a ticket here: https://youtrack.jetbrains.com/issue/KT-45443

soywiz:
Thanks @SerVB !

B8EE1FEC-C16A-4AF8-9003-D4B84D3A9B62

PR #194, opened 2021-03-08
Add mappings for test to ViewsForTesting.sceneTest…

serandel:
...to override the module with stub, mocks, etc.

My previous version of the method was cumbersome to use if I wanted to override something from the module, so I expanded it :)

### DIFF
--- a/korge/src/commonMain/kotlin/com/soywiz/korge/tests/ViewsForTesting.kt
+++ b/korge/src/commonMain/kotlin/com/soywiz/korge/tests/ViewsForTesting.kt
@@ -13,6 +13,7 @@ import com.soywiz.korge.internal.*
 import com.soywiz.korge.scene.*
 import com.soywiz.korge.view.*
 import com.soywiz.korgw.*
+import com.soywiz.korinject.AsyncInjector
 import com.soywiz.korio.async.*
 import com.soywiz.korio.lang.*
 import com.soywiz.korio.util.*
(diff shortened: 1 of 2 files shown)


korge/src/commonMain/kotlin/com/soywiz/korge/tests/ViewsForTesting.kt
```diff
@@ -222,8 +223,9 @@ open class ViewsForTesting(
 	}
 
     @Suppress("UNCHECKED_CAST")
-    inline fun <reified S : Scene> sceneTest(
+    inline fun <reified S : Scene> ViewsForTesting.sceneTest(
```
soywiz, line 226: Why is this receiver required?

korge/src/commonTest/kotlin/com/soywiz/korge/ViewsForTestingTest.kt
```diff
@@ -4,31 +4,49 @@ import com.soywiz.korge.scene.Module
...
         }
     }
 
+    object DummyModule : Module() {
+        override suspend fun AsyncInjector.configure() {
+            mapSingleton {
+                Dependency()
+            }
+            mapSingleton {
+                DummyScene(get())
```
soywiz, line 30: Scenes are intended to be prototypes instead of singletons, since everytime you change a scene a new instance is created. Even if this is a test, this could be misleading. Could you change it?

soywiz (commented):
I think the new lambda makes sense. 👍

Could you check my inline comments?

korge/src/commonMain/kotlin/com/soywiz/korge/tests/ViewsForTesting.kt
serandel: Copy and paste mistake from my extension function, sorry.

korge/src/commonTest/kotlin/com/soywiz/korge/ViewsForTestingTest.kt
serandel: Yeah, I'm scoping the modules so in my game they're singletons, but I definitely see your point, changing to Prototype.

soywiz:
Thanks @serandel !

dumb_scene

serandel:
I'll keep doing PRs if only to see your memes at the end 😂😂😂

PR #202, opened 2021-03-29
Add DropshadowFilter

soywiz:


soywiz:
Thanks, Carlos!

dropshadow

PR #203, opened 2021-03-30
Fast sprite improvements - support for rotation and color

LeHaine:
**Context:**
`FastSprite` and `FastSpriteContainer` didn't seem to support setting colors/alpha and rotation. 

**Changes**:
I added support to allow optional rotation calculation for `FastSprite` and `FastSpriteContainer`. It is disabled by default. When it is disabled it uses the previous `BatchBuilder2D` draw call in `FastSpriteContainer`. If it is enabled, `FastSprite` will calculate the vertices based on the rotation.

**Tests**:
I was getting 40 FPS on the JVM target with 200k bunnies before any changes (without rotation). With rotation enabled I am also getting 40 FPS on JVM. with 200k bunnies. Would like to see someone else try this out and see if they get similar results.

### DIFF
--- a/korge/src/commonMain/kotlin/com/soywiz/korge/view/fast/FastSprite.kt
+++ b/korge/src/commonMain/kotlin/com/soywiz/korge/view/fast/FastSprite.kt
@@ -27,6 +27,11 @@ open class FastSprite(tex: BmpSlice) {
     // sin(rotation) value
     var sr: Float = 0f
 
+    internal var useRotation = false
+
+    /**
+     * Updates based on rotation
+     */
     private fun updateXY0123() {
         val px = xf - ax
         val py = yf - ay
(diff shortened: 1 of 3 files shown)


soywiz:
Which computer are you using? Maybe it is related to CPU cache, since the sprite object is now bigger

LeHaine:
I think I had the vertices rotation calculations slightly wrong which I just corrected. 

I am using an AMD Ryzen 5 3600 with GeForce RTX 2060 Super GPU. Some additional tests:

200K bunnies with `fastSpriteContainer(useRotation = true)`
java_ZC9EIYLRie

200K bunnies with `fastSpriteContainer(useRotation = false)`
java_XFy21I9fcT

200K bunnies using `korge-next` master branch
java_UUv6XTLg8Q

LeHaine:
I think I still have calculation issues with rotation. I'll need to spend some more time on it. I'll convert this to a draft in the mean time.

LeHaine:
Calculations are reworked and the rotation works and rotates correctly around the anchor points as it should now. Did the same test as above and received similar results either way. But this is ready for any review.

soywiz:
Thanks! Really appreciated, this allows new use-cases. Thanks @LeHaine !

rotated_bunny

PR #210, opened 2021-04-15
Fix left whitespaces in headers

SerVB:
### Tests fail:

* Whitespace before the value isn't ignored: `Headers((Connection, [ close]), (Content-Length, [ 349]), (Content-Type, [ text/html]), (Date, [ Thu, 15 Apr 2021 20:08:51 GMT]), (Server, [ ECSF (dcb/7EC7)]))`
* Expected true: `assertTrue(UrlVfs("https://example.com").open().readAll().isNotEmpty())`

### Tests pass:

* `assertTrue(UrlVfs("https://example.com").readAll().isNotEmpty())`

## After this PR

All these tests pass.

## Fixed issues

As discussed [here](https://discord.com/channels/728582275884908604/801658049244430377/831272197133172766), playing mp3 files taken from UrlVfs fails with `IllegalStateException: Not an mp3 file [2]`. It's because under the hood, `open().readAll()` was invoked that returned an empty array. After this PR, code like `UrlVfs("https://file-examples-com.github.io/uploads/2017/11/file_example_MP3_700KB.mp3").readAudioStream(MP3Decoder).playAndWait()` works.

However, this fix doesn't solve unsupported streaming playing, so code like `UrlVfs("https://air2.radiorecord.ru:805/rr_320").readAudioStream(MP3Decoder).playAndWait()` will continue throwing `IllegalStateException: Not an mp3 file [2]`. I've created korlibs/korge#729 about that.

## Rationale

As described at <https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers>, "Whitespace before the value is ignored". It wasn't true for korio before this PR. This caused some issues afterwards. For example, UrlVfs stat size was evaluated to 0 due to `"<space>number".toLongOrNull()` gives `null`, not `number`:

https://github.com/korlibs/korge-next/blob/15d618b6d771b87ef788f4a859c9f548e7c0f0e7/korio/src/commonMain/kotlin/com/soywiz/korio/file/std/UrlVfs.kt#L147

So I've decided to add trimming left white spaces in headers' values as proposed by the standard.

### DIFF
--- a/korio/src/commonMain/kotlin/com/soywiz/korio/net/http/HttpPortable.kt
+++ b/korio/src/commonMain/kotlin/com/soywiz/korio/net/http/HttpPortable.kt
@@ -10,7 +10,11 @@ import kotlinx.coroutines.*
 import kotlin.coroutines.*
 import kotlin.math.*
 
-internal object HttpPortable {
+open class HttpPortable(
+    private val factory: AsyncSocketFactory = asyncSocketFactory
+) {
+    companion object : HttpPortable()
+
     internal fun computeHeader(method: Http.Method, url: URL, rheaders2: Http.Headers): String = buildString {
         val EOL = "\r\n"
         append("$method ${url.pathWithQuery} HTTP/1.1$EOL")
@@ -26,7 +30,7 @@ internal object HttpPortable {
 				val url = URL(url)
 				val secure = url.scheme == "https"
 				//println("HTTP CLIENT: host=${url.host}, port=${url.port}, secure=$secure")
-				val client = createTcpClient(url.host!!, url.port, secure)
+				val client = factory.createClient(url.host!!, url.port, secure)
 
 				val rheaders = combineHeadersForHost(headers, url.host)
 				val rheaders2 = if (content != null)
@@ -87,7 +91,7 @@ internal object HttpPortable {
 
 			override suspend fun listenInternal(port: Int, host: String) {
 				val context = coroutineContext
-				val socket = createTcpServer(port, host)
+				val socket = factory.createServer(port, host)
 				actualPort = socket.port
 				val close = socket.listen { client ->
 					while (true) {
--- a/korio/src/commonMain/kotlin/com/soywiz/korio/net/http/HttpPortable.kt
+++ b/korio/src/commonMain/kotlin/com/soywiz/korio/net/http/HttpPortable.kt
@@ -45,7 +45,7 @@ open class HttpPortable(
 				val firstLine = client.readLine()
 				val responseInfo = Regex("HTTP/1.\\d+ (\\d+) (.*)").find(firstLine) ?: error("Invalid HTTP response $firstLine")
 
-				//println("FIRST LINE: $firstLine")
+				//println("FIRST LINE: ${firstLine.trim()}")
 
 				val responseCode = responseInfo.groupValues[1].toInt()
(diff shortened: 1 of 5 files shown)


korio/src/commonMain/kotlin/com/soywiz/korio/net/http/HttpPortable.kt
```diff
@@ -56,7 +56,7 @@ internal object HttpPortable {
 
 				val responseHeaders = Http.Headers(headers.map {
 					val parts = it.split(':', limit = 2)
-					parts.getOrElse(0) { "" } to parts.getOrElse(1) { "" }
+					parts.getOrElse(0) { "" } to parts.getOrElse(1) { "" }.trimStart()
```
soywiz, line 59: 👍

korio/src/commonTest/kotlin/com/soywiz/korio/file/std/UrlVfsTest.kt
```diff
@@ -1,10 +1,17 @@
...
 
 class UrlVfsTest {
     @Test
     fun test() {
         assertEquals("http://127.0.0.1:8080/", UrlVfs("http://127.0.0.1:8080")["."].absolutePath)
     }
+    @Test
+    fun testUrlVfsRead() = suspendTest {
+        assertTrue(UrlVfs("https://example.com").readAll().isNotEmpty())
+        assertTrue(UrlVfs("https://example.com").open().readAll().isNotEmpty())
```
soywiz, line 15: I don't like the idea on relying on external services for testing. This source set covers fast unit/integration tests but not end2end. We should use something mocked or open a temporal local server.

Let me check how to do mocked requests with korio 👀 and I will come back to you

korio/src/commonTest/kotlin/com/soywiz/korio/net/http/HttpHeadersTest.kt
```diff
@@ -8,4 +9,19 @@ class HttpHeadersTest {
...
     }
+    @Test
+    fun testIgnoredWhitespaceInHeadersValue() = suspendTest {
+        // this test checks "Whitespace before the value is ignored" rule
+        // from https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers
+
+        val fullUrl = "https://example.com"
+        val client = HttpPortable.createClient()
+
+        val result = client.request(Http.Method.HEAD, fullUrl)
```
soywiz, line 20: Same here. Let me check how to test this without relying on external services

soywiz (commented):
Let's figure out a way to test the http client without actually performing external connections

soywiz:
Figured out. I have done this in a branch. Can you backport those changes to your PR, so I can merge it while giving you credit for the fix?

* https://github.com/korlibs/korge-next/commit/b0d916477e7cb21fae7894f68beb351c004cf9df - Make it testable
* https://github.com/korlibs/korge-next/commit/3cb820512450b4b667728fa330281d73dcf1a785 - Testing using a `FakeAsyncClient`

SerVB:
Hey @soywiz, I've removed my tests and cherry-picked your commits. So it seems the only changed line by me here is where I add `.trimStart()` 😅 But it's totally OK for me, feel free to merge.

soywiz:
Everything's fine. Thanks! :)

PR #211, opened 2021-04-16
Creates fromUnix function to return with applied timezone

Caps-Looking:
Fix https://github.com/korlibs/klock/issues/158

### DIFF
--- a/klock/src/commonMain/kotlin/com/soywiz/klock/DateTimeTz.kt
+++ b/klock/src/commonMain/kotlin/com/soywiz/klock/DateTimeTz.kt
@@ -24,6 +24,12 @@ class DateTimeTz private constructor(
         /** Creates a new local [DateTimeTz] from a [unix] time */
         fun fromUnixLocal(unix: Double): DateTimeTz = DateTime(unix).localUnadjusted
 
+        /** Creates a new local [DateTimeTz] from a [unix] time applied*/
+        fun fromUnix(unix: Long): DateTimeTz {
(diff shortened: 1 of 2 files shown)


klock/src/commonTest/kotlin/com/soywiz/klock/TimeFormatTest.kt
```diff
@@ -53,4 +53,15 @@ class TimeFormatTest {
         assertEquals(59, time.second)
         assertEquals(0, time.millisecond)
     }
+
+    @Test
+    fun testFromUnix() {
```
soywiz, line 58: 👍 no needed to mock the timezone in the end right?

klock/src/commonMain/kotlin/com/soywiz/klock/DateTimeTz.kt
```diff
@@ -24,6 +24,12 @@ class DateTimeTz private constructor(
         /** Creates a new local [DateTimeTz] from a [unix] time */
         fun fromUnixLocal(unix: Double): DateTimeTz = DateTime(unix).localUnadjusted
```
soywiz, line 25: Do you think it makes sense to deprecate this? Could it be confusing to other people?

Caps-Looking: Absolutely!

soywiz:
Sorry for the huge delay. I totally forgot about this. Thanks!

PR #212, opened 2021-04-16
Fix 3d

SerVB:
I want to try 3d but I see cubes aren't rendered (Collada models are rendered OK). I've debugged a bit and see that vertex count for cubes is 0, maybe it's the reason. I've just added a test for this (it fails currently), I will try to fix this vertex count later (so please don't merge this PR yet).

Related: korlibs/korge#306, korlibs/korge-samples#35.

Maybe @dhakehurst who reworked 3D (#35) can suggest something here. Maybe also somebody else can comment. I'm a total noob in 3d and even don't understand most of terminology, so it can be quite difficult for me to understand the problem.

### DIFF
--- a/korge/src/commonMain/kotlin/com/soywiz/korge3d/Shapes3D.kt
+++ b/korge/src/commonMain/kotlin/com/soywiz/korge3d/Shapes3D.kt
@@ -74,13 +74,15 @@ class Cube3D(var width: Double, var height: Double, var depth: Double) : ViewWit
                     val dx = dirs[0]
                     val dy = dirs[1]
 
-                    addVertex(pos - dx - dy, normal, Vector3D(0f, 0f, 0f))
-                    addVertex(pos + dx - dy, normal, Vector3D(1f, 0f, 0f))
-                    addVertex(pos - dx + dy, normal, Vector3D(0f, 1f, 0f))
+                    val i0 = addVertex(pos - dx - dy, normal, Vector3D(0f, 0f, 0f))
+                    val i1 = addVertex(pos + dx - dy, normal, Vector3D(1f, 0f, 0f))
+                    val i2 = addVertex(pos - dx + dy, normal, Vector3D(0f, 1f, 0f))
 
-                    addVertex(pos - dx + dy, normal, Vector3D(0f, 1f, 0f))
-                    addVertex(pos + dx - dy, normal, Vector3D(1f, 0f, 0f))
-                    addVertex(pos + dx + dy, normal, Vector3D(1f, 1f, 0f))
+                    val i3 = addVertex(pos - dx + dy, normal, Vector3D(0f, 1f, 0f))
+                    val i4 = addVertex(pos + dx - dy, normal, Vector3D(1f, 0f, 0f))
+                    val i5 = addVertex(pos + dx + dy, normal, Vector3D(1f, 1f, 0f))
+
+                    addIndices(i0, i1, i2, i3, i4, i5)
                 }
 
                 face(Vector3D(0f, +.5f, 0f))
--- a/korge/src/commonTest/kotlin/com/soywiz/korge3d/MeshTest.kt
+++ b/korge/src/commonTest/kotlin/com/soywiz/korge3d/MeshTest.kt
@@ -1,0 +1,13 @@
+package com.soywiz.korge3d
+
+import kotlin.test.*
+
+@OptIn(Korge3DExperimental::class)
+class MeshTest {
+    @Test
+    fun testCube() {
+        val cube = Cube3D(1.0, 1.0, 1.0)
+        val vertexCount = cube.mesh.vertexCount
+        assertTrue(vertexCount > 0, "cube.mesh.vertexCount(=$vertexCount) is not positive")
+    }
+}


SerVB:
Oh, it seems I've just figured it out. There is a mesh without indices for cube. Fortunately, it's not related to internals. I've added indices to cube mesh and now boxes from the demo are visible. @dhakehurst , sorry for bothering, but maybe you can approve the idea.

Until this fix comes to a release and for older versions, the **workaround** is to copy cube to project's sources, fix the issue there, and use it instead of cube from KorGE:

<details>

<summary>Code to paste to your project</summary>

```kotlin
internal val vector3DTemps = Vector3DTemps()

internal class Vector3DTemps {
    @PublishedApi
    internal var pos = 0
    @PublishedApi
    internal val items = arrayListOf<Vector3D>(Vector3D(), Vector3D(), Vector3D())

    fun alloc(): Vector3D {
        val npos = pos++
        return if (npos < items.size) {
            items[npos]
        } else {
            val item = Vector3D()
            items.add(item)
            item
        }
    }

    inline operator fun <T> invoke(callback: Vector3DTemps.() -> T): T {
        val oldPos = pos
        try {
            return callback()
        } finally {
            pos = oldPos
        }
    }

    operator fun Vector3D.plus(that: Vector3D) = alloc().setToFunc { this[it] + that[it] }
    operator fun Vector3D.minus(that: Vector3D) = alloc().setToFunc { this[it] - that[it] }
}

@Korge3DExperimental
inline fun Container3D.cube(width: Int, height: Int, depth: Int, callback: Cube3D.() -> Unit = {}): Cube3D = cube(width.toDouble(), height.toDouble(), depth.toDouble(), callback)

@Korge3DExperimental
inline fun Container3D.cube(
    width: Double = 1.0,
    height: Double = width,
    depth: Double = height,
    callback: Cube3D.() -> Unit = {}
): Cube3D = Cube3D(width, height, depth).addTo(this, callback)

@Korge3DExperimental
class Cube3D(var width: Double, var height: Double, var depth: Double) : ViewWithMesh3D(mesh.copy()) {
    override fun prepareExtraModelMatrix(mat: Matrix3D) {
        mat.identity().scale(width, height, depth)
    }

    var material: Material3D?
        get() = mesh.material
        set(value) {
            mesh.material = value
        }

    fun material(material: Material3D?): Cube3D {
        this.material = material
        return this
    }

    companion object {
        val mesh = MeshBuilder3D {
            vector3DTemps {
                fun face(pos: Vector3D) {
                    val dims = (0 until 3).filter { pos[it] == 0f }
                    val normal = Vector3D().setToFunc { if (pos[it] != 0f) 1f else 0f }
                    val dirs = Array(2) { dim -> Vector3D().setToFunc { if (it == dims[dim]) .5f else 0f } }
                    val dx = dirs[0]
                    val dy = dirs[1]

                    val i0 = addVertex(pos - dx - dy, normal, Vector3D(0f, 0f, 0f))
                    val i1 = addVertex(pos + dx - dy, normal, Vector3D(1f, 0f, 0f))
                    val i2 = addVertex(pos - dx + dy, normal, Vector3D(0f, 1f, 0f))

                    val i3 = addVertex(pos - dx + dy, normal, Vector3D(0f, 1f, 0f))
                    val i4 = addVertex(pos + dx - dy, normal, Vector3D(1f, 0f, 0f))
                    val i5 = addVertex(pos + dx + dy, normal, Vector3D(1f, 1f, 0f))

                    addIndices(i0, i1, i2, i3, i4, i5)
                }

                face(Vector3D(0f, +.5f, 0f))
                face(Vector3D(0f, -.5f, 0f))

                face(Vector3D(+.5f, 0f, 0f))
                face(Vector3D(-.5f, 0f, 0f))

                face(Vector3D(0f, 0f, +.5f))
                face(Vector3D(0f, 0f, -.5f))
            }
        }
    }
}
```

</details>

soywiz:
Awesome. Thank you very much!

dhakehurst:
Strange it does not work...I thought I had tests and examples for cubes.

The issue is likely in MeshBuilder3D. cuboid, though that call faceRectangle, which does add indices.


ah....Cube3D is not using the cube or cuboid method from MeshBuilder3D....that is probably the real issue here.

dhakehurst:
ah yes....I always just use shape3D, which give access to the methods on MeshBuilder3D
e.g.

```
shape3D {
    cube(2f)
}
```

thus Cube3D and the cube method are really redundant, or at least should just do the same as above as a convenience thing.

SerVB:
@soywiz , so we now can create cubes as `cube(w, h, d)` and `shape3D { cuboid(w, h, d) }` in KorGE. Moreover, they have different implementations. Do you think it's better to delegate from the first to the second (thus dropping one of the implementations)? Maybe even remove the first one (of course, deprecate it firstly)?

soywiz:
Yeah, agree with that. Not checked, but I guess the option with more overhead should use the one with less overhead / allocations etc. If not possible because they are incompatible or require allocations, then create a low overhead generator of the shape, and then make the two of the implementations use that one. Something like that I guess

dhakehurst:
my plan was to deprecate cube(w, h, d) or make it call shape3D { cuboid(w, h, d) }

PR #214, opened 2021-04-20
Bumps Kotlin to 1.5.0-RC

soywiz:
Closes https://github.com/korlibs/korge-next/pull/183


korgw/src/jvmMain/kotlin/com/soywiz/korgw/win32/Win32Tools.kt
```diff
@@ -46,7 +46,14 @@ interface Win32GL : INativeGL, Library {
     fun glGenVertexArrays(n: Int, out: IntArray)
     fun glBindVertexArray(varray: Int)
 
-    companion object : Win32GL by Win32GL.OpenglLoadProxy() {
```
soywiz, line 49: This produced a null pointer exception that caused an initialization error. So this used to work and with IR 1.5.0-RC it fails

PR #217, opened 2021-04-22
Add Layer, FrameTag and Slice config to AtlasInfo

jobe-m:
In Aseprite it is possible to save in the sprite atlas config file
also information about layers, frame tags and slices. This commit
adds support for reading this additional information also into the
AtlasInfo object. If this information is not present in a json
file then the three lists "layers", "frameTags" and "slices" will
be empty.

I uses the "slices" feature of Aseprite to define additional regions
in a sprite. This is very handy if you want to paint these regions
within one image. Particularly I use this feature for a parallax effect
for the background of my game. I hope I can make an example
when I have some more time.

I added with this commit also the layers and frame tags information
into the AtlasInfo class. So it should comes in handy when you need
this additional information in the game/sprite setup.

My editor has reformatted some lines. There are no changes done in
those.


soywiz:
@jobe-m nice! Could you add a sample file, and maybe a test loading it so I can review it with context?

jobe-m:
Would you fancy if I push my current game status. It's noch much yet, just the parallax background. But it wouldn't compile without the slices added to AtlasInfo in a local build 2.0.0.999. So you could use it as a test :)

image

jobe-m:
You can find my game repo here: https://github.com/jobe-m/jobesLegacyKorge

The slices from AtlasInfo are used here: https://github.com/jobe-m/jobesLegacyKorge/blob/df6bce72fb7b12d92065135cb4a6c83c46785e67/src/commonMain/kotlin/image/ParallaxBackground.kt#L81

And a json file with the additional info you can find here: https://github.com/jobe-m/jobesLegacyKorge/blob/master/src/commonMain/resources/world01level01/parallax_background.json

soywiz:
Nice!

As I can see, it seems to provide frames and duration information:
https://github.com/jobe-m/jobesLegacyKorge/blob/df6bce72fb7b12d92065135cb4a6c83c46785e67/src/commonMain/resources/world01level01/parallax_background.json#L8

That could be used to create [SpriteAnimation](https://github.com/korlibs/korge-next/blob/d939ae6f99faa911a3a9ecbaf31b07e6a60428f2/korge/src/commonMain/kotlin/com/soywiz/korge/view/SpriteAnimation.kt) instances somehow. But maybe in a separate PR.

Regarding to the sample added to the PR, you could add it to the `korim/src/commonTest/resources` folder: https://github.com/korlibs/korge-next/tree/d939ae6f99faa911a3a9ecbaf31b07e6a60428f2/korim/src/commonTest/resources

If those graphics are yours and you don't mind the be available here, you can upload those. If not, maybe you can create a small file with dummy graphics in Aseprite or even upload that json with a blank png file of that size, and then add a test trying to load that resource. If you know meaningful assets to perform would be nice, if not at least trying to load it, we know there are no exceptions thrown

jobe-m:
Good, Idea. I could look into adding the duration information to a SpriteAnimation object. Also constructing a SpriteAnimation from frameTags with same name / pattern would be good. I will check that in another PR.

Having a test as you mentioned is a good idea, too. I like meaningful tests :)
I will create some template files (I wanted to do them anyway as explanation how the ParallaxBackground can be used). Those files I can add with a test to Korge. I will look into that.

jobe-m:
I hope the test setup is correct and that the test is started/run by your test framework :)

soywiz:
Nice. Thanks!

PR #219, opened 2021-04-24
[WIP] Instanced rendering + float textures + support for multiple separated vertex buffers / layouts

soywiz:
Also reimplemented bunnymark in the `instanced-rendering` sample using an experimental `FSprites` that stores each sprite packed so now bunnymark is more GPU-bound instead of CPU/memory bound. Can achieve 800K sprites in bunnymark on my cpu/gpu at 60fps in a 800x600 buffer / 110 fps if reducing the window size.


soywiz:
I will split this work in several PRs

PR #228, opened 2021-04-27
fix for korio userHomeVfs to be path in System.getProperty("user.home")

dhakehurst:
fix for korio userHomeVfs to be path in System.getProperty("user.home")


soywiz:
Thanks!

PR #229, opened 2021-04-29
Feature/kotlin.native.inline.optimizations2

soywiz:
Some of these commits hurt JVM at least. We have to check. At least FSprite could benefit from the first two commits I guess without hurting JVM.

> Each function call on Kotlin/Native / propery with custom setter/getter is calling `EnterFrame/LeaveFrame` which is specially expensive on Windows since it uses a thread local and writes/reads memory even if the function is arithmetic only, and also prevents inlining and cross-function optimizations when linking. The trick tried here is to try to inline as much as possible stuff so at least critical places are inlined, optimized and reduced as much as possible the EnterFrame/LeaveFrame issue.

Details here: https://kotlinlang.slack.com/archives/C3SGXARS6/p1619349974244300


soywiz:
Not needed with Kotlin 1.5.0 (only needed manually inlining const val to circumvent this issue https://youtrack.jetbrains.com/issue/KT-46425 ) already on master

PR #230, opened 2021-04-29
Bump kotlin to 1.5.0

soywiz:


soywiz:
I believe it fails because some static initialization order issue, so something changed from JVM Legacy to IR t here

PR #232, opened 2021-04-29
Add unit test for watch function in resourcesVfs

jobe-m:
The added "watch" unit test should pass since the watch call of
resourcesVfs is inside a launch lambda. Currently it fails because of a
bug in LocalVfsJvm.kt (most probably...).

This is for reproducing issue korlibs/korge#727

### DIFF
--- a/korio/src/jvmMain/kotlin/com/soywiz/korio/file/std/LocalVfsJvm.kt
+++ b/korio/src/jvmMain/kotlin/com/soywiz/korio/file/std/LocalVfsJvm.kt
@@ -402,22 +402,16 @@ private class LocalVfsJvm : LocalVfsV2() {
 		val fs = FileSystems.getDefault()
 		val watcher = fs.newWatchService()
 
-		fs.getPath(path).register(
+        val registeredKey = fs.getPath(path).register(
 			watcher,
 			StandardWatchEventKinds.ENTRY_CREATE,
 			StandardWatchEventKinds.ENTRY_DELETE,
 			StandardWatchEventKinds.ENTRY_MODIFY
 		)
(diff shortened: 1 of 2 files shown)


soywiz:
That test seems to prevent gradle to finish

jobe-m:
Yes, exactly. This test is for reproducing the issue mentioned above.
Interestingly the "watch" function is called inside "launch" and the test is still suspending the current thread.
I thing that the watch functionality is somehow handlwed wrongly. When I have time I will try to figure this behaviour.

korio/src/jvmMain/kotlin/com/soywiz/korio/file/std/LocalVfsJvm.kt
```diff
@@ -402,22 +402,16 @@ private class LocalVfsJvm : LocalVfsV2() {
...
-					var r: WatchKey?
-					do {
-						r = watcher.poll(100L, TimeUnit.MILLISECONDS)
-					} while (r == null && running)
-					r
-				} ?: continue
+                val key = watcher.take()
 
 				for (e in key.pollEvents()) {
 					val kind = e.kind()
```
soywiz, line 417: Could you wrap `when (kind) { ... }` with `withContext(coroutineContext) { when (kind) { ... } }` that way handler will be executed in the right thread where the function was called, in korio and korge this is important since it is mostly single threaded and mostly not thread-safe

soywiz (approved):

soywiz:
LGTM. Thanks! 

korio_watching

jobe-m:
Thanks for merging :)

PR #234, opened 2021-05-01
Fix/touch.event.fixes

### LINKED ISSUE #161: Fix click not working properly on some targets

### LINKED ISSUE #193: 2.0.7.0 breaks clicks on mobile device browser in my game
Found out that for some time clicks don't work when I open my game in a mobile web browser.

I see errors in console, seems like e.touches is empty (on this line: https://github.com/korlibs/korge-next/blob/498c12a1cd80272b8673ea2e736ad72410369389/korge/src/commonMain/kotlin/com/soywiz/korge/Korge.kt#L373)

image

I've investigated a bit: checked git history and saw this PR: https://github.com/korlibs/korge-next/pull/162

It seems that it spoils touches: on 2.0.7.0 (with this PR merged), the bug persists; on 2.0.6.0 (before this PR), clicks work fine.

To reproduce, you can build my game. Here it has KorGE 2.0.7.0: https://github.com/SerVB/pph/blob/e478eff981571ec5bd4370010ceb03f947b6ea5b/gradle.properties#L1. If you access the built game via Mobile Chrome, clicks will produce exceptions. If you change the version to 2.0.6.0, clicks will work fine.

### LINKED ISSUE #205: Multi-touch detection with input instance not recording
KorGE version: 2.0.7.1
OS: Windows

Looks like the input instance isn't recording touch correctly for the activeTouches list, as Console consistently shows the list is empty in this code block:

*For context, levelData is a Container object. It is added as a child to a scene. 

```kotlin
       var nTouches: Int

        levelData.addUpdaterWithViews { views, dt ->
            nTouches = views.input.activeTouches.size

            Console.log("activeTouches size: $nTouches")

            if (nTouches == 2) {
                testPlayer.nullify()
            }
            else {
                testPlayer.reactivate()
            }
        }
```
image

I'll be implementing a workaround or a different motion for the feature. Should be reproduce-able through a new KorGE project with this code block: 
```kotlin
	var nTouches: Int

	this.addUpdaterWithViews { views, dt ->
		nTouches = views.input.activeTouches.size

		Console.log("activeTouches size: $nTouches")
	}
```

Let me know if any more details are required!

### LINKED ISSUE #772: Android device: onClick event is about previous click
Developing an app based on https://github.com/korlibs/korge-hello-world I noticed that sometimes touching a view on a physical Android device doesn't work (not shown in onClick event of the view), 
Then I found out that next touch of another view results in onClick event of previously touched view. 
And so on: on a touch previous event appears.
The same problem on my Android 7 and 11 physical devices.

This problem doesn't occur on Android emulators (and on JVM), where I use PC's mouse to "touch" that views.

The problem is easily reproduced and traced using statements like this:

    Console.log("Logo clicked")

I understand that korlibs v.1.* on which current version of korge-hello-world is based, aren't fixed anymore, so I'm waiting for the korge-hello-world to be updated to use KorGe 2.0, so I will reproduce (or not) the problem there.

For the reference, my app is this: https://github.com/andstatus/game2048 , the same problem is reproduced in the sample app https://github.com/RezMike/2048

soywiz:
* Fixes https://github.com/korlibs/korge-next/issues/205
* Fixes https://github.com/korlibs/korge-next/issues/161
* Fixes https://github.com/korlibs/korge/issues/772
* Fixes https://github.com/korlibs/korge-next/issues/193


korgw/src/commonMain/kotlin/com/soywiz/korev/Events.kt
```diff
@@ -82,11 +90,100 @@ data class Touch(
...
     override fun hashCode(): Int = index
     override fun equals(other: Any?): Boolean = other is Touch && this.index == other.index
+
+    fun toStringNice() = "Touch[${id}][${status}](${x.niceStr},${y.niceStr})"
+}
+
+// On JS: each event contains the active down touches (ontouchend simply don't include the touch that has been removed)
+// On Android: ...
+// On iOS: each event contains partial touches with things that have changed for that specific event
+class TouchBuilder {
```
SerVB, line 107: Is there any reason to have all the implementations in the common source set? For now, I guess implementations can be moved to platform source sets via expect/actual, so generated binaries won't contain redundant implementations.

soywiz: I didn't want to use `JS, Android, IOS` as part of the enum, but something like `PROVIDE_EVENT_TOUCHES`, `PROVIDE_ACTIVE_TOUCHES`, `PROVIDE_ALL_TOUCHES`, so it is more like a behaviour of which touches are provided in each event more than the platform itself, that could be reused for several platforms. But since I didn't come with good names yesterday, I named them like that lol 🤣. This API shouldn't be used by most people so we can change it.
Android is not using TouchBuilder IIRC, so maybe I can provide only the two implementations, or check how are android providing touches and using one of the two or add the other one. In the end I believe implementations could provide touches on each even in those three ways, and this tries to normalize that

PR #236, opened 2021-05-02
Fixed compile error for NativeNativeSoundProvider.kt on Linux

jfbilodeau:
Fixed compile error for NativeNativeSoundProvider.kt on Linux due to Kotlin 1.5 outlawing Double -> Short conversion
    https://youtrack.jetbrains.com/issue/KT-30360


soywiz:
Thanks!

PR #240, opened 2021-05-03
Added FastChar and FastCharArray and use it in StrReader.readStringLit

soywiz:
Temporal workaround for https://youtrack.jetbrains.com/issue/KT-46443


soywiz:
Not required anymore as for Kotlin 1.6.20 where Char is already inline, and CharArray is backed by UInt16Array.

PR #258, opened 2021-05-30
Add staggered isometric tilemap support

ObsidianX:
Supports:
 - X/Y Stagger Axis
 - Odd/Even Stagger Index
image


soywiz:
Awesome. Thanks!

PR #259, opened 2021-05-31
korge issue #367: Update supportBox2d() commit ID

ObsidianX:
Fixes https://github.com/korlibs/korge/issues/367

Updated to point to https://github.com/korlibs/korge-bundles/commit/7439e5c7de7442f2cd33a1944846d44aea31af0a which references kbox2d 2.1.1

Added a log when loading a bundle to print the SHA256 sum when one isn't provided to simplify updating in the future.

### DIFF
--- a/korge-gradle-plugin/src/main/kotlin/com/soywiz/korge/gradle/KorgeExtension.kt
+++ b/korge-gradle-plugin/src/main/kotlin/com/soywiz/korge/gradle/KorgeExtension.kt
@@ -406,7 +406,7 @@ class KorgeExtension(val project: Project) {
 
     fun supportBox2d() {
         // https://awesome.korge.org/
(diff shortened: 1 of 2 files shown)


soywiz:
Thanks!

It seems that on the sample we used the bundle directly: https://github.com/korlibs/korge-samples/blob/b3f007ee30c51c827bb9443898b778778a6d1c50/samples/box2d/build.gradle.kts#L14

TobseF:
I tried to use Box-2D in hallo world from the Korge wizard.
The gradle sync works. But the import for `import org.jbox2d.dynamics.BodyType` is not available.

I tried both:
```kotlin
supportBox2d()
```
and the long version:
```kotlin
bundle("https://github.com/korlibs/korge-bundles.git::korge-box2d::7439e5c7de7442f2cd33a1944846d44aea31af0a##9fd9d54abd8abc4736fd3439f0904141d9b6a26e9e2f1e1f8e2ed10c51f490fd")
```

Anything else to configure?
korgePluginVersion=`2.1.1.4`

soywiz:
Uhm. That's strange. I'll check  👀

PR #260, opened 2021-05-31
Further staggered isometric tilemap improvements

ObsidianX:
- Apply the current transformation matrix to the newly defined dimensions to scale properly
- Correctly determine `initY` based on the difference between the map's tile height and the tileset's tile height
- Use the map's tile height to determine the stagger offset
- ~Switch blending mode to premultiplied alpha `(ONE, ONE_MINUS_SRC_ALPHA)` to eliminate black borders on transparent edges in staggered isometric tile textures~ (ref: https://gamedev.stackexchange.com/a/21202)

### DIFF
--- a/korge/src/commonMain/kotlin/com/soywiz/korge/view/tiles/TileMap.kt
+++ b/korge/src/commonMain/kotlin/com/soywiz/korge/view/tiles/TileMap.kt
@@ -110,26 +110,26 @@ open class TileMap(
 
         val renderTilesCounter = ctx.stats.counter("renderedTiles")
 
-        val initY = (if (staggerAxis != null) tileSize.height - tileHeight else 0.0).let {
-            min2(m.transformX(it, 0.0), m.transformY(0.0, it))
-        }
         val posX = m.transformX(0.0, 0.0)
         val posY = m.transformY(0.0, 0.0)
         val dUX = m.transformX(tileWidth, 0.0) - posX
         val dUY = m.transformY(tileWidth, 0.0) - posY
         val dVX = m.transformX(0.0, tileHeight) - posX
         val dVY = m.transformY(0.0, tileHeight) - posY
+        val initY = (if (staggerAxis != null) tileSize.height - tileHeight else 0.0).let {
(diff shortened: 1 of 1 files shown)


korge/src/commonMain/kotlin/com/soywiz/korge/view/tiles/TileMap.kt
```diff
@@ -311,7 +326,7 @@ open class TileMap(
 
         infos.fastForEach { buffer ->
             ctx.batch.drawVertices(
-                buffer.vertices, ctx.getTex(buffer.tex), smoothing, renderBlendMode.factors, buffer.vcount, buffer.icount
+                buffer.vertices, ctx.getTex(buffer.tex), smoothing, AG.Blending(AG.BlendFactor.ONE, AG.BlendFactor.ONE_MINUS_SOURCE_ALPHA), buffer.vcount, buffer.icount
```
soywiz, line 329: If this is not working, this should be a separate issue. Maybe related to this? https://github.com/korlibs/korge/issues/95

That's usually related to premultiplied alpha. I have found in some cases PNG images that had alpha premultiplied against the specification. There are ways to asume it as premultiplied alpha like here: https://github.com/korlibs/korge-samples/blob/b3f007ee30c51c827bb9443898b778778a6d1c50/samples/spine/src/commonMain/kotlin/mainCommon.kt#L12

Also calling `AG.Blending` performs an allocation, so at least it should be cached. But I believe it should be handled elsewhere.

Can you remove this line for this PR and provide a tilemap/tileset reproducing your issue so we can investigate?

ObsidianX: Sounds good! I'll put together a sample project with the tileset to share.

soywiz: Thanks! Once we have a sample that we can use, we could also add tests verifying that everything works well (including ortogonal, staggered and plain isometric when supported), so we can refactor safely later

soywiz:
Thanks!

PR #266, opened 2021-06-29
Use a default framerate of 60fps if the actual value fails to load

ObsidianX:
On Linux the X11GraphicsDevice implementation will return a refresh rate of 0 if it couldn't get the current display mode for whatever reason.


soywiz:
Nice. Thanks!

PR #271, opened 2021-07-09
Clean up types used in JNA mapping of libGL using OpenGL type names

ObsidianX:
Also added the ability to override libopenal.so from the builtin natives in KorAU.  The linuxarm copy does not work on Raspberry Pi but instead of making it a Raspberry Pi compatible version just override it at runtime.


korgw/src/jvmMain/kotlin/com/soywiz/korgw/platform/INativeGL.kt
```diff
@@ -2,164 +2,187 @@ package com.soywiz.korgw.platform
 
 import com.soywiz.kmem.FBuffer
 import com.soywiz.korim.bitmap.NativeImage
+import com.sun.jna.NativeLong
 import java.nio.ByteBuffer
 import java.nio.FloatBuffer
 import java.nio.IntBuffer
 
-typealias IntSize = Long
+typealias GLboolean = Boolean
```
soywiz, line 10: Due to JNA behavior, this can't be a bool. We had a bug related to this with some drivers on linux. Explanation here:
https://github.com/korlibs/korge/issues/268#issuecomment-729056184

soywiz: I believe it should be Byte, and explicitly use 1 or 0

soywiz:
Going to merge and fix the GLboolean issue in a separate commit. Thanks!

PR #315, opened 2021-07-23
Supports freeing and tinting FSprites

soywiz:


korge/src/commonMain/kotlin/com/soywiz/korge/view/fast/FSprites.kt
```diff
@@ -168,6 +198,8 @@ inline class FSprite(val id: Int) {
     //val offset get() = index * STRIDE
 }
 
+fun FSpriteFromIndex(index: Int) = FSprite(index * FSPRITES_STRIDE)
```
LeHaine, line 201: Is this supposed to be an extension function by chance for `FSprites`?

soywiz: Woops. Forgot to remote it. The problem is that FSprite constructor uses not the index of the sprite, but the stride for the original array, initially because of performance, but it is a bit confusing. I think I should change that so the constructor is the index.

PR #316, opened 2021-07-23
Raspberry Pi support

ObsidianX:
Initial support for the `linuxArm32Hfp` (Raspberry Pi) build target.  The entire target is behind the new `enableKotlinRaspberryPi` flag in `gradle.properties` which is disabled by default.

A few of the `linuxX64` source files were usable without modification but I haven't found a way to share the source between the two versions yet.  Building the Pi version does require a local sysroot (/tmp/rpi-sysroot) folder to properly link with the ARM binaries.  The `cinterop` target doesn't support setting linker options in Gradle however `binary` does which would help remove the hard-coded path to the sysroot, but that would be a larger change.

`mimalloc` causes segfaults on the Raspberry Pi and the `useMimalloc` flag is ignored for the `linuxArm32Hfp` target.


ObsidianX:
One note before this can be merged: The `korge-editor` module needs to be disabled for `linuxArm32Hfp()` but I'm not sure how to do that.

gradle.properties
```diff
@@ -3,7 +3,7 @@ version=2.0.0.999
 #kotlinVersion=1.5.20
 kotlinVersion=1.5.30-M1
 kotlinSerializationVersion=1.2.1
-coroutinesVersion=1.5.0
+coroutinesVersion=1.5.1-SNAPSHOT
```
ObsidianX, line 6: I just remembered this relies on a custom build of kotlinx.coroutines with Arm32Hfp enabled.  Probably can't merge this unless that gets enabled upstream.

soywiz: I'm going to merge it and disable it with a flag or something, so the code is already here and we can enable it when this situation changes.

Did you do something special to compile kotlinx.coroutines & atomicfu on the PI?

soywiz:
pi-korge

Thank you very much!

gradle.properties
ObsidianX: For kotlinx.coroutines I needed to disable tests on the arm target: https://gist.github.com/ObsidianX/dcfb862a6f41fad598053eef50679374

soywiz: Because some of the tests failed, or something else?

ObsidianX: Posted a PR with a better solution: https://github.com/Kotlin/kotlinx.coroutines/pull/2841

Issue #329, opened 2021-08-06
Do not draw hidden layers of ASE file


jobe-m:
This is related to #314

I read through the ASE file specification and now have a good understanding what needs to be done in ASE.kt file for ignoring the hidden/invisible layers of an ASE file.

As first approach we will remove all hidden layers from the image data after the ASE file was read.
It looks like that we need to check each cell of a frame if it is located on a invisible (hidden) layer. If yes, than that cell should be dropped from the final image data.
Then we create a new list of all visible layers and add that also to the image data.
Finally the (self pointing) index of all visible layers needs to be adjusted because if we remove invisible layers than the list of visible layers is shorter. That means that list might contain layers which have a wrong index. So we will need to adjust that index, too. For that a change from "val index" to "var index" in ImageLayer class is needed.

PR #330, opened 2021-08-06
Remove invisible layers from image data of ASE file

jobe-m:
Each cell of a frame is checked if it is located on an invisible (hidden)
layer. If yes, than that cell is dropped from the final image data. This
effectively means that no image data from any invisible layer is
available for drawing in the image data.

A new list of all visible layers is created and added to the image data.
The indexes of all visible layers are adjusted so that all entries have
the correct index. That is needed in ImageAnimationView class for
referencing the correct layer object. For that a change from "val index"
to "var index" in ImageLayer class was needed.

ASETest.kt was updated with a new ase file which contains
hidden/invisible layers. In order to test also a ase file with slices
the existing slice-example.ase was updated with a second invisible
layer. Both files are read correctly and the images are taken into the
generated atlas.


jobe-m:
This fixes #329

soywiz:
Thanks!

hidden_layer

PR #337, opened 2021-08-22
Add SDL2 support for Linux targets

ObsidianX:
Using SDL allows you to run KorGE outside of an X11 environment on Linux which removes a lot of overhead on low powered devices like the Raspberry Pi Zero.  When running in the framebuffer through SDL, KorGE appears to run near 60FPS compared to an estimated 10-15FPS when running natively through X11.

To use, set:
`KORGW_JVM_ENGINE=sdl` when running in JVM
`KORGW_NATIVE_ENGINE=sdl` when running in Native

When running in JVM you can use an alternate libSDL2.so with:
`SDL2_PATH=/usr/local/lib/libSDL2.so`


korgw/src/commonMain/kotlin/com/soywiz/korgw/sdl2/SDLKeyCode.kt
```diff
@@ -0,0 +1,270 @@
...
+    EJECT(SDLScancode.EJECT.keycode),
+    SLEEP(SDLScancode.SLEEP.keycode),
+    APP1(SDLScancode.APP1.keycode),
+    APP2(SDLScancode.APP2.keycode),
+
+    AUDIOREWIND(SDLScancode.AUDIOREWIND.keycode),
+    AUDIOFASTFORWARD(SDLScancode.AUDIOFASTFORWARD.keycode);
+
+    companion object {
+        private val lookup = mutableMapOf<Int, SDLKeyCode>()
```
soywiz, line 262: values().associatedBy { it.value }

korgw/src/jvmMain/kotlin/com/soywiz/korgw/sdl2/jna/enums/SDL_KeyCode.kt
```diff
@@ -0,0 +1,270 @@
...
+    EJECT(SDL_Scancode.EJECT.keycode),
+    SLEEP(SDL_Scancode.SLEEP.keycode),
+    APP1(SDL_Scancode.APP1.keycode),
+    APP2(SDL_Scancode.APP2.keycode),
+
+    AUDIOREWIND(SDL_Scancode.AUDIOREWIND.keycode),
+    AUDIOFASTFORWARD(SDL_Scancode.AUDIOFASTFORWARD.keycode);
+
+    companion object {
+        private val lookup = mutableMapOf<Int, SDL_KeyCode>()
```
soywiz, line 262: Ditto

korgw/src/jvmMain/kotlin/com/soywiz/korgw/sdl2/jna/enums/SDL_WindowEventID.kt
```diff
@@ -0,0 +1,33 @@
...
+    ENTER,
+    LEAVE,
+    FOCUS_GAINED,
+    FOCUS_LOST,
+    CLOSE,
+    TAKE_FOCUS,
+    HIT_TEST;
+
+    companion object {
+        val lookup = mutableMapOf<Int, SDL_WindowEventID>()
```
soywiz, line 25: Ditto

korgw/build.gradle.kts
```diff
@@ -28,8 +28,14 @@ if (doEnableKotlinNative) {
     kotlin {
         for (target in nativeTargets()) {
             target.compilations["main"].cinterops {
-                if (target.name == "linuxX64") maybeCreate("GL")
-                if (target.name == "linuxArm32Hfp") maybeCreate("GL_rpi")
+                if (target.name == "linuxX64") {
+                    maybeCreate("GL")
```
soywiz, line 32: Would this force people to have SDL even if they use the other backend?

soywiz (approved):
5910201A-25C4-4D90-BFCD-3355E0F50CF3

Awesome! I think we should allow the user to decide which backend to use at compile time. But we can do that later

Issue #339, opened 2021-08-22
Create artifact variants for linux SDL and non-SDL


soywiz:
So people can select the backend used without requiring it at runtime.

There are other alternatives here:
* Use dlopen on linux to dynamically use SDL .h files are not required at compiletime
* Implement direct framebuffer on linux to not require SDL for raspi zero

@ObsidianX

PR #342, opened 2021-09-05
Rewrite SDL implementation to load dynamically

ObsidianX:
Removes the requirement of having SDL development libraries installed locally for each target platform.  Cleaned up all the extra "-sdl" "-nosdl" folders.


korge/src/commonMain/kotlin/com/soywiz/korge/render/BatchBuilder2D.kt
```diff
@@ -870,5 +870,5 @@ class BatchBuilder2D constructor(
 }
 
 //internal const val BB_MAX_TEXTURES = 6
-internal const val BB_MAX_TEXTURES = 4
+internal const val BB_MAX_TEXTURES = 1
```
soywiz, line 873: We should probably configure this dynamically. This would prevent other targets from using multiple textures

soywiz:
Fixes https://github.com/korlibs/korge-next/issues/339

soywiz:
Going to merge it and fix some stuff

sdl

Thanks!

korge/src/commonMain/kotlin/com/soywiz/korge/render/BatchBuilder2D.kt
ObsidianX: Oh dang it, this was supposed to be reverted for this pr, sorry!

soywiz: No problem. No need to revert anymore: https://github.com/korlibs/korge-next/blob/8e16ee712ceea65866e18aa0597f2fe5a94de358/korge/src/commonMain/kotlin/com/soywiz/korge/render/BatchBuilder2D.kt#L876-L879

PR #346, opened 2021-09-09
Fix AM/PM parsing bug when hour is 12, and ignore case of AM/PM in parsed string

CharlesKozel:
Fix AM/PM parsing bug when hour is 12, and ignore case of AM/PM in parsed string.


soywiz:
Woops. It seems I totally forgot about reviewing this. Thank you very much!

am_pm

PR #347, opened 2021-09-17
Set linuxArm32Hfp to use 1 texture batch as well

ObsidianX:


soywiz:
Woops. I believe I added the wrong target in the first place :)

thanks!

PR #348, opened 2021-09-18
Update GL.def

conifer-dev:
Include Solus linux distro lib location to the linker.


soywiz:
new_path

Awesome. Thank you!

PR #354, opened 2021-09-23
Added Dyn.toStringOrNull()

jfbilodeau:


korio/src/commonMain/kotlin/com/soywiz/korio/dynamic/Dyn.kt
```diff
@@ -98,6 +98,14 @@ inline class Dyn(val value: Any?) : Comparable<Dyn> {
 
     override fun toString(): String = toString(value)
 
+    fun toStringOrNull(): String? {
+        return if (this.isNotNull) {
```
soywiz, line 102: Probably not what you want: You are comparing the Dyn instance, that is always going to be non-null AFAIK. Probably that this, should be `value`.

Also, can you add a test here?
https://github.com/korlibs/korge-next/blob/ae2a5ed06acc3c240d9a6b24d0be78870c4f0886/korio/src/jvmTest/kotlin/com/soywiz/korio/dynamic/DynTest.kt#L55

jfbilodeau:
Thanks for the feedback. The `isNotNull()` function is defined in `Dyn` as follows:
```kotlin
 val isNotNull get() = value != null
```
so it does check value.

I'll add a test.

soywiz:
Woops :) you are right. Thanks for the addition!

PR #391, opened 2021-11-11
Add parallax data container and view with example code

jobe-m:
This commit adds an example which shows how to use the new
ParallaxDataView class to create a nice parallax effect. Also all
needed changed and additions to Korge/Korim are done in a backwards
compatible manner. Hopefully! I checked the korge-starter-kit-rpg
when it still was in this repo and it worked ;)

Take your time to review and let me know if something should be
done differently.

The following is a summary for all changes:

RepeatedImageView was added which implements drawing an image which
shall repeats in X and Y direction. This code was taken/inspired by
TileMap class. RepeatedImageView is used for drawing the layer images
which normally will repeat over the screen.

AbstractImageAnimationView class was added. That is used to be able
to choose two different implementations for ImageAnimationView in
ImageDataView class. In ImageAnimationView class the layers property
was changed from "fastArrayList" of Image to "mutableMap" of string
to Image. This was needed because the parallax Data view needs to be
able to access the layers of the Aseprite image by its name.

ImageDataView class and its inline function was extended to support
also repeating the image object. Thus depending of the "repeating"
value different implementations are used for "standard" images
(-> imageAnimationView) and "repeating" images
(-> repeatingImageAnimationView). This was done since I did not
manage to implement "repeating" in Image class in the first run.

The RepeatedImageAnimationView is in priciple a copy of the
ImageAnimationView class where the Image object (representing the
layers) was exchanged with RepeatedImageView.

~~In ImageFrameLayer the targetX and targetY properties were set to
mutable ("var") because it is needed to be able to set these values
after reading the ASE file.~~ Not, needed it was left over from my first
tries. I just realized it while looking into the failing unit test ASETest.

The following is done in ASE.kt (readImageContainer)
In case a sliced image is read from the
ASE file then the target position is set regarding the relative
position of the slice in the Aseprite image. The slices are used for
the parallax plane where each slice is a stripe of the plane and
therefore the original position of the stripe needs to be taken over
from the Aseprite image. (See ImageDecodingProps.useSlicePosition in
ASE.kt file.)

In ParallaxDataView class the parallax drawing logic is implemented.
It gets the data for the parallax layers and the parallax plane from
the ParallaxDataContainer. That is an object which stores different
Image and ImageContainer objects. They are read all from one Aseprite
file. For controlling what data is read from the Aseprite file three
ImageDecodingProps are used.

In file ASE.kt the function readImageContainer was updated to be able
to selectively load specific layers by name. For that
ImageDecodingProps "layers" is used. If that is set then only layers
are read which name matches the "layers" prop list. It is also
possible to disable slicing of the image with "disableSlicing" prop.
With that it is possible to selectively slice only selected layers.
Finally there is another ImageDecodingProps "useSlicePosition". With
that it is possible to set the position of a sliced image object
relative to the original image. This is useful if the sliced image
should be placed relative to another sliced image in the view which
uses these images.


soywiz:
WOW. First of all, thank you very much for this awesome contribution and all the work. This looks really remarkable.

I'll try to review this as soon as I can, either tomorrow or during the weekend. 👀

jobe-m:
I will update the ASETest shortly so that the test should pass :)

korge/src/commonMain/kotlin/com/soywiz/korge/view/RepeatedImageView.kt
```diff
@@ -0,0 +1,206 @@
...
+
+inline fun Container.repeatedImageView(
+    bitmap: BmpSlice,
+    repeatX: Boolean = false,
+    repeatY: Boolean = false,
+    smoothing: Boolean = true,
+    block: @ViewDslMarker RepeatedImageView.() -> Unit = {}
+) = RepeatedImageView(bitmap, smoothing, repeatX, repeatY).addTo(this, block)
+
+open class RepeatedImageView(
```
soywiz, line 25: Do you think RepeatedImageView could be replaced by a TileMap with a single cell + a single image TileSet like in this sample?

https://github.com/korlibs/korge-samples/blob/231fa10e28e44a0e1415aaa1d031c11f382a55bd/samples/tiled-background/src/commonMain/kotlin/main.kt#L11-L12

korge/src/commonMain/kotlin/com/soywiz/korge/view/animation/ImageAnimationView.kt
```diff
@@ -24,25 +41,28 @@ open class ImageAnimationView(animation: ImageAnimation? = null, direction: Imag
     var direction: ImageAnimation.Direction? = direction
 
     private val computedDirection: ImageAnimation.Direction get() = direction ?: animation?.direction ?: ImageAnimation.Direction.FORWARD
-    private val layers = fastArrayListOf<Image>()
+    private val layers = mutableMapOf<String, Image>()
```
soywiz, line 44: Could it be possible to keep the layers list, while still having a `layersByName` using a FastStringMap?
At least for iteration this would be benefical, specially in JS

korge/src/commonMain/kotlin/com/soywiz/korge/view/animation/ParallaxDataView.kt
```diff
@@ -0,0 +1,432 @@
...
+                }
+            }
+        }
+
+        // Last construct the foreground layers on top
+        constructLayer(data.foregroundLayers, data.config.foregroundLayers, data.config.mode, smoothing)
+    }
+}
+
+suspend fun VfsFile.readParallaxDataContainerFromAseFile(
```
soywiz, line 278: Could it be renamed to `readParallaxDataContainer` and pass the file format as parameter? That way it could be used from example from a `.psd` or a `.kra` file or a custom format provided by someone using other program

korge/src/commonMain/kotlin/com/soywiz/korge/view/animation/ParallaxDataView.kt
```diff
@@ -0,0 +1,432 @@
...
+    config: ParallaxConfig,
+    atlas: MutableAtlas<Unit>? = null
+): ParallaxDataContainer {
+    val props = ImageDecodingProps(this.baseName, extra = ExtraTypeCreate())
+    return ParallaxDataContainer(
+        config = config,
+        backgroundLayers = if (config.backgroundLayers != null) {
+            props.setExtra("layers", config.backgroundLayers.joinToString(separator = ",") { it.name })
+            props.setExtra("disableSlicing", true)
+            val out = ASE.readImage(this.readAsSyncStream(), props)
```
soywiz, line 288: This line:

```
val out = ASE.readImage(this.readAsSyncStream(), props)
```

seems to be repeated a few times. It is reading the same ASE file?

samples/parallax-scrolling-aseprite/src/main/AndroidManifest.xml
```diff
@@ -0,0 +1,3 @@
+<?xml version="1.0" encoding="utf-8"?>
+<manifest package="com.soywiz.korge.parallaxscrolling.aseprite">
```
soywiz, line 2: Was this file needed on samples?

korge/src/commonMain/kotlin/com/soywiz/korge/view/RepeatedImageView.kt
jobe-m: Yes, I guess this would be possible. In fact I created RepeatedImageView.kt by making a copy of TileMap.kt and throwing out everything what I did not need.

korge/src/commonMain/kotlin/com/soywiz/korge/view/animation/ImageAnimationView.kt
jobe-m: I think that can be done.

korge/src/commonMain/kotlin/com/soywiz/korge/view/animation/ParallaxDataView.kt
jobe-m: Yes, that's also fine.

korge/src/commonMain/kotlin/com/soywiz/korge/view/animation/ParallaxDataView.kt
jobe-m: Yes, it reads the same ASE file. The reason is that I need at least once read the file without "slicing" enabled (see "disableSlicing" ImageDecodingProps) and once read the file with slicing enabled. Slicing is used for generating the parallax (3D) plane from the ASE file. The normal layer images (e.g. for the foreground layer) should not be sliced from the ASE file but just taken as the full image.
Then I was also easier handling the different layer types (background, foreground and attached layers) in the ParallaxDataView when they are in different ImageData objects. This is why I read it 4 times.

samples/parallax-scrolling-aseprite/src/main/AndroidManifest.xml
jobe-m: I just copy-pasted it. Don't remember from where actually ;)  I can be removed I guess.

korge/src/commonMain/kotlin/com/soywiz/korge/view/animation/ImageAnimationView.kt
jobe-m: Done :)

korge/src/commonMain/kotlin/com/soywiz/korge/view/RepeatedImageView.kt
jobe-m: I just tried to use a TileMap and now I remember why I did the RepeatedImageView as a "strip down" of a TileMap. (That was quite some time ago...) It is impossible to change the image inside the TileSet without recreating the whole object. But that is needed in setFrame function. So it would be not very efficient to have animated images for the layers of the parallax background (which I do not yet have but which might be an option for animated backgrounds :)

korge/src/commonMain/kotlin/com/soywiz/korge/view/animation/ParallaxDataView.kt
jobe-m: Done :)

samples/parallax-scrolling-aseprite/src/main/AndroidManifest.xml
jobe-m: File removed

korge/src/commonMain/kotlin/com/soywiz/korge/view/animation/ParallaxDataView.kt
```diff
@@ -275,7 +275,8 @@ class ParallaxDataView(
     }
 }
 
-suspend fun VfsFile.readParallaxDataContainerFromAseFile(
+suspend fun VfsFile.readParallaxDataContainer(
+    formats: ImageFormat = RegisteredImageFormats,
     config: ParallaxConfig,
```
soywiz: Since it is optional, maybe would be nice to move it after config. And since `ImageFormat` is singular, maybe the parameter could be renamed to `format`

korge/src/commonMain/kotlin/com/soywiz/korge/view/animation/ParallaxDataView.kt
soywiz: I see. Thanks for the explanation!

korge/src/commonMain/kotlin/com/soywiz/korge/view/RepeatedImageView.kt
soywiz: I see. Maybe we can create an abstract class or extract the functionality to some internal methods so the code is not duplicated.
Not high priority, and we can skip this for now for this PR, but would be nice to have at some point to avoid those two implementations evolving differently or not improving both of them when updating one.
What do you think?

soywiz:
Had to do the initial review on the iPad, but jus tested on the PC a few minutes ago and the demo is really cool, congratulations! :)

korge/src/commonMain/kotlin/com/soywiz/korge/view/RepeatedImageView.kt
jobe-m: I fully agree. I can check how to "merge" RepeatedImageView with TileMap class for that.

Another question: I looked into TileMap for "repeating" the Bitmap because that was already implemented there (for the tiles). I did not succeed to add "repeating" of the Bitmap in e.g. BaseImage when I first looked into it. Which is quite some while ago. Now I have a bit better understanding of the Korge sources on that topic.
What do you think. Should we try to add "repeatX" and "repeatY" to BaseImage? Or should the "repeat" feature be kept separate in the TileMap class?

korge/src/commonMain/kotlin/com/soywiz/korge/view/animation/ParallaxDataView.kt
jobe-m: You are right. I even can set it to "ASE" instead to "RegisteredImageFormats".

korim/src/commonMain/kotlin/com/soywiz/korim/format/ImageData.kt
```diff
@@ -28,15 +28,15 @@ open class ImageData constructor(
     val animationsByName = animations.associateBy { it.name }
     val area: Int get() = frames.area
     val framesByName = frames.associateBy { it.name }
-    val framesSortedByProority = frames.sortedByDescending {
+    val framesSortedByPriority = frames.sortedByDescending {
```
soywiz, line 31: Woops :) Thanks for noticing!

korge/src/commonMain/kotlin/com/soywiz/korge/view/RepeatedImageView.kt
jobe-m: I just looked into the code again and have an idea how I can merge RepeatedImageView back to TileMap. I have the idea to create an abstract base class "BaseTileMap" and derive from that TileMap and something like "SingleTile" (as replacement for RepeatedImageView).

While looking into that I realized that RepeatedImageAnimationView can also be "merged" back to ImageAnimationView class. There I would also save a lot duplicated code.

I have time to look into it this evening...

soywiz: Cool. That's a nice idea!

jobe-m:
Looks like there is still one test failing but I don't understand why...

soywiz:
@jobe-m thats a timeout, probably unrelated, let me rerun it

soywiz (approved):
Awesome 🎊 thank you very mucho for the contribution!

PR #395, opened 2021-11-21
Fixed NPE in ParticleEmitterView.kt when changing scene

jfbilodeau:
I get an NPE in ParticleEmitterView when changing scene. It seems that the `stage` becomes null while the particle emitter is running, leading to an NPE. This PR should solve it.

### DIFF
--- a/korge/src/commonMain/kotlin/com/soywiz/korge/particle/ParticleEmitterRead.kt
+++ b/korge/src/commonMain/kotlin/com/soywiz/korge/particle/ParticleEmitterRead.kt
@@ -69,6 +69,7 @@ suspend fun VfsFile.readParticleEmitter(): ParticleEmitter {
         }
     }
 
+//    emitter.textureName = null
     emitter.texture = try {
         file.parent[emitter.textureName?.takeIf { it.isNotBlank() } ?: "texture.png"].readBitmapSlice()
     } catch (e: FileNotFoundException) {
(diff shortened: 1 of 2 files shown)


soywiz:
Woops. It is strange that this happens on an `addUpdater` since it is being executed when attached to the stage.
I noticed that the comment and the old change is reverted, this fixed another issue. Is that issue not happening at this point?
Is by chance easily reproducible adding a test to figure out it is fixed?

jfbilodeau:
> Woops. It is strange that this happens on an `addUpdater` since it is being executed when attached to the stage. I noticed that the comment and the old change is reverted, this fixed another issue. Is that issue not happening at this point? Is by chance easily reproducible adding a test to figure out it is fixed?

Oups! The reverted change is a boo boo. Sorry for that. I'll see if I can get a mock up to demo the NPE.

jfbilodeau:
I'm going to close this request since I'm providing the fix in another PR

soywiz:
@jfbilodeau Thanks for the fix. Can you reopen this one, and remove the fix in the other PR so we can handle each issue in separate PRs? That will be better for git history

jfbilodeau:
Sure...sorry for the really stupid question, but do you know how to remove a file from a PR? Sorry for my lack of Git knowledge... :P

soywiz:
No problem at all. I can help you if you want with this, can you ping me via discord?

PR #397, opened 2021-11-22
Make addUpdater configurable in ParallaxDataView

jobe-m:
(I saw that you are currently on vacation. Don't worry, this PR can wait.
Please relax, recharge your batteries and have great holidays.)

The ParallaxDataView will add an "updater" to each of its view
which scrolls depending to the game playfield. Sometimes the parallax
view should not scroll automatically but the position should be set
directly. E.g. when using parallax backgrounds in an intro or similar
where it is needed to have more "control" over the movement of the
background.

Thus this commit makes adding the "addUpdater" lambda configurable per X
and Y scrolling direction.


soywiz:
Thanks!

jobe-m:
You absolutely got it with this animation :)

PR #398, opened 2021-11-28
Implement storing bitmap font into atlas

jobe-m:
The readBitmapFont inline function and the BitmapFont class were
extended with a parameter for a mutual atlas object. If an atlas
object is given than the letters of the Bitmap font are stored in the
atlas. Otherwise the BitmapFont class behaves as before.

An unit test was also added with two example bitmap fonts in txt and
xml format.


soywiz:
Super nice. Thanks!

its_own_texture

PR #400, opened 2021-12-01
Fix parent dir path error in TiledMaps

### LINKED ISSUE #399: Cannot load tiledmaps where Tilesets are not in the same folder as the Tiledmap (relative path issue)
**Describe the bug**

Using Korge next (2.4.8 / 2.0.0.999 and Kotlin 1.6), Korge cannot load a .tmx TiledMap if it references .tsx TileSets stored in a different directory than the .tmx. For example, given this .tmx XML:

```xml
 <tileset firstgid="1" source="32x32_map_tile.tsx"/>
 <tileset firstgid="668" source="../dungeon/Dungeon.tsx"/>
```
The first .tsx file loads correctly, but the second fails with the following error:

`java.io.FileNotFoundException: D:\Development\Games\Korge\TheAlchemist\src\commonMain\resources\maps\village\dungeon\Dungeon.tsx (The system cannot find the path specified)`

The correct path should be:  `D:\Development\Games\Korge\TheAlchemist\src\commonMain\resources\maps\dungeon\Dungeon.tsx`
I'm guessing the trouble is somewhere in korio/file but I haven't been able to track it down.

**Annoyment level**

> How annoyed are you in a scale of 1-3 because this shitty code doesn't work properly

* [x] 1 - No rush from my side
* [ ] 2 - It is inconvenient, but I can circumvent the issue for now
* [ ] 3 - FUUUU
* [ ] 4

**Payment**

> How do you plan to compensate the time we will invest fixing this?

* [x] I will add a message with "Thanks" when merged to let you know I'm grateful
* [ ] I will also add a [meme image](https://imgflip.com/memegenerator) relevant to the issue to make you smile so we both have fun
* [ ] In fact, I'm going to add a meme here in this issue too
* [ ] I'm going to write about KorGE in my blog, spread the voice on social networks, or maybe even make a TikTok wearing a KorGE T-Shirt while dancing

> You can also consider donating some money via Github Sponsors to the person fixing the issue.

**Meme**

> Here you can put a meme to motivate us to fix the issue.

jobe-m:
There was a problem in tilemap (tmx) files with reading a tileset (tsx)
file when the tileset file is located outside of the directory where the
tilemap file is located. It looks like that the TiledMapData object is
storing the folder path in a jailVfs. This leads to the problem that all
additional files of that TiledMapData object needs to be located inside
of the same folder and subfolders. Removing the "jail" fixed the above
mentioned problem.

An unit test was added which reproduces the initial problem.

This should fix #399


soywiz:
Awesome. Thanks! Maybe the jail should be provided from outside in the case it is required for security reasons.

jail

jobe-m:
@soywiz You are right. I will check if with this current solution we can escape out of the resources folder. That would be then a security flaw. Good idea to limit the jail to the resources folder instead.

jobe-m:
I just checked if it is possible to break out of the resourceVfs by moving the `tsx` file out and putting in `demo.tmx` line 3 `../../wood_tileset_3.tsx`. 

https://github.com/korlibs/korge-next/blob/090a68a83aba83c392d1f514ab66bf69255b2960/korge/src/commonTest/resources/tiled/demo.tmx#L3

But the error log of the TileMapTest shows that Korge still builds the path inside of the `resouces` folder. So all fine :)

```
com.soywiz.korge.tiled.TiledMapTest[jvm] > testTileMapWithTileSetFromOutsideFolder[jvm] FAILED
    java.io.FileNotFoundException: /home/jobe/devel/korge-next/korge/src/jvmMain/resources/wood_tileset_3.tsx (No such file or directory)
```

soywiz:
Nice. Thanks for checking @jobe-m !

I believe your use case is perfectly reasonable. When I implemented it I thought that the map and the tileset was going to be in the same folder. But didn't thought on that case.
The resources vfs is jailed or should be jailed. The same should happen for other filesystems. If not, we could by default provide them jailed and provide unjailed versions like: currentFolderVfs and currentFolderVfsUnjailed.
In any case is up to the developer to decide what to jail and what not to jail, but we should make it easy to be as safe as possible when makes sense.

But again, I believe I did a wrong assumption when implementing this initially and the solution you proposed is the right one :) thanks again!

PR #402, opened 2021-12-14
Linux sdl show simple message box and fix NPE

jfbilodeau:
Swtiched from Zenity to SDL_ShowSimpleMesageBox
(This commit accidentally includes a fix for NPE with Particle System on changing scene)


korge/src/commonMain/kotlin/com/soywiz/korge/particle/ParticleEmitterView.kt
```diff
@@ -100,8 +100,8 @@ class ParticleEmitterView(
 //                simulator.emitterPos.setTo(x, y)
 //            }
             if (dt > 0.milliseconds) {
-                val gx = globalX / stage!!.scaleX
-                val gy = globalY / stage!!.scaleY
+                val gx = globalX / (stage?.scaleX ?: 1.0)
+                val gy = globalY / (stage?.scaleY ?: 1.0)
```
soywiz, line 104: This seems from the previous PR, can you revert these changes here, so we can merge this PR?

PR #403, opened 2021-12-14
Fix updating frame index in ImageAnimationView

jobe-m:
For getting nextFrameIndex the prior index is saved in frameIndex and
not in "frame.index" which is the number of the frame within ImageData
frame list.

I found this while having two longer animations in my ase file. The second
animation was not played properly. It looked that the frames are jumping
in random order. You don't realize this effect in the korge-starter-kit-rpg sprites
since the animations for the sprites are only 3 frames long.


soywiz:
Is it possible somehow to add a test for this, to avoid future regressions when refactoring?

jobe-m:
> Cool, thanks for noticing and fixing @jobe-m ! :)
> 
> Is it possible somehow to add a test for this, to avoid future regressions when refactoring?

Good idea, I will look into it.

soywiz (approved):
Nice. Thanks! Feel free to merge 👍

jobe-m:
I will merge it now.
The unit test currently cannot check if the frames are animated in the correct order since the setFrame function is private. Also, I don't want to change it to protected since private functions are an internal implementation detail which might change in the future. When adding new functionality to the ImageAnimationView class I will check to implement it in a way which makes it better testable.

PR #412, opened 2021-12-26
Fix weekOfYear having more than 7 days

sdoward:
As per ISO-8601 the first week of the year starts on the first Monday of the year. Any days before that count towards the last week of the previous year.e.g.

Screenshot 2021-12-26 at 17 16 48

The current implementation assigns all of those days to week 0. Meaning that week 0 can have up to 13 days (I think).

**This PR**

I created a couple of test which reproduce the issue. 

The implementation uses recursion to solve the issue. I don't know if it is the nicest approach but it fixes the problem.

**Future Work**

This Api could be improved as some developers might be caught out that days in January are assigned to week 52 or 53. I feel like a `YearWeek` class makes sense. It would be a similar API to `YearMonth`


soywiz:
Thanks @sdoward !

I'm checking other implementations:

* https://momentjs.com/docs/#/get-set/week/
* https://momentjs.com/docs/#/get-set/iso-week/
* https://en.wikipedia.org/wiki/ISO_week_date

It seems that there are two typical week of year implementations.
I would do the same here:

```
moment("12-26-2021").week()
1
moment("12-26-2021").isoWeek()
51
```

Can we do the same here, so it doesn't affect the original implementation?

sdoward:
@soywiz Thanks for the quick response. I don't follow exactly... "12-26-2021" is never week 1 is it?

soywiz:
@sdoward  It seems that in moment.js week of the year is 1-based instead of 0-based

sdoward:
FYI I am using this website to help me visualise https://www.epochconverter.com/weeks/2021

PR #413, opened 2021-12-29
Add test case for datetime formatting

sdoward:
I believe this to be a bug but I am not 100% sure.

I create a unix timestamp for 01-01-2020 00:00:00 and then format it. The result is off by 12 hours. 

 `expected:<2020-01-01 [00]:00:00> but was:<2020-01-01 [12]:00:00>`

### DIFF
--- a/klock/src/commonTest/kotlin/com/soywiz/klock/DateFormatTest.kt
+++ b/klock/src/commonTest/kotlin/com/soywiz/klock/DateFormatTest.kt
@@ -8,4 +8,12 @@ class DateFormatTest {
         assertEquals("Sat, 03 Dec 2011 10:15:30 GMT+0100", DateFormat.FORMAT1.parse("2011-12-03T10:15:30+01:00").toStringDefault())
         assertEquals("Sat, 03 Dec 2011 10:15:30 GMT+0100", DateFormat.FORMAT1.parse("2011-12-03T10:15:30+0100").toStringDefault())
     }
+
+    @Test
+    fun testTimeStampFormatting() {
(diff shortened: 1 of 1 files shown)


sdoward:
Closing, happy to report I was using H

soywiz:
Hey @sdoward , thanks for the PR!
Just checked and I believe the result is expected. I think the problem is that `hh` lowercase puts hours in 12-hour format (1-12) instead of 24-hour format (0-23).
The 24-hour version is `HH` uppercase.
This is a bit misleading because minutes `mm` and seconds `ss` are lowercase. But it was added like that because was the original format from java: https://docs.oracle.com/javase/7/docs/api/java/text/SimpleDateFormat.html

https://github.com/korlibs/korge-next/blob/25d4520f3636657e12c1639aea03b9c7cd3eb716/klock/src/commonMain/kotlin/com/soywiz/klock/PatternDateFormat.kt#L152-L156

Can you verify if changing `YYYY-MM-dd hh:mm:ss` to `YYYY-MM-dd HH:mm:ss` provides an expected value for you?

sdoward:
@soywiz Thanks for getting back to me.

I don't think that assumption is correct. Midnight on a 12 and 24 hour clock should be the same. They are both 00:00.

If I use `HH` it formats correctly. Using `hh` produces a different time

sdoward:
Yes correct. Thanks for looking into this

soywiz:
Nice! Glad it helped.

Closing the issue, feel free to open a new one if you find something strange.

PR #415, opened 2022-01-02
Corrected transform order in VertexShader

Stechmuck:
This corrects the transform order in the VertexShader in korlibs/korge#560


jobe-m:
Looks good to me 👍

soywiz:
Thanks for the fix @Stechmuck ! And thanks for checking @jobe-m !
LGTM too.

We cannot merge it yet because the CI is failing and that would break master.
I believe it fails in a reference test, so it should be easily fixable. You should be able to fix it by executing:

`./gradlew jvmTestFix`

Then comiting the updated files (probably `FSprites1.log`, `FSprites2.log`, etc.)

Stechmuck:
I have updated the files

soywiz:
Awesome. Thank you very much @Stechmuck !

PR #423, opened 2022-01-09
Add Base64url variant of Base64 encoding

gsteckman:
[Base64url](https://datatracker.ietf.org/doc/html/rfc3548#page-6) is a slight variation of Base64 encoding that uses only URL-safe characters and can optionally not use padding. This is an implementation of the Base64url variant leveraging the existing Base64 codec and taking the approach described in [RFC 7515 Appendix C](https://datatracker.ietf.org/doc/html/rfc7515#appendix-C).


krypto/src/commonMain/kotlin/com/soywiz/krypto/encoding/Base64url.kt
```diff
@@ -0,0 +1,41 @@
...
+            2 -> s += "=="
+            3 -> s += "="
+            else -> throw IllegalArgumentException("Illegal base64url string")
+        }
+
+        return Base64.decode(s)
+    }
+
+    fun decodeIgnoringSpaces(str: String): ByteArray {
+        return decode(str.replace(" ", "").replace("\n", "").replace("\r", "").replace("\ts", ""))
```
soywiz, line 23: Probably \ts should be \t

krypto/src/commonMain/kotlin/com/soywiz/krypto/encoding/Base64url.kt
```diff
@@ -0,0 +1,41 @@
...
+    }
+
+    fun encode(src: ByteArray, doPadding: Boolean = false): String {
+        return Base64.encode(src).replace("+", "-")
+            .replace("/", "_").let {
+                if (!doPadding) {
+                    it.split("=")[0]
```
soywiz, line 30: I believe we can use substringBefore instead or split to have better performance

soywiz (commented):
Nice addition. Thanks!
Could you check my comments?

Also, is the url version a superset in terms of characters used?
Could we strip the padding by default so the decode always work within the original implementation/extension member, and then provide an optional parameter for encoding for the url? The objective would be to simplify the API and maybe only have the Base64 object, and some extra methods/parameters there

krypto/src/commonMain/kotlin/com/soywiz/krypto/encoding/Base64url.kt
gsteckman: I copied this directly from Base64.decodeIgnoreingSpaces, where it also has "\ts", and didn't notice this. But I agree it should probably be "\t", in both places.

soywiz: Oh fu :) then we had a bug, and something to test

gsteckman:
The url version essentially replaces "+" with "-" and "/" with "_" in the allowable alphabet, and allows no padding to be used in cases where the data length is otherwise known, whereas I believe technically Base64 requires padding although people often leave it off.

I can think of two alternative API approaches that would leave only the Base64 object:

1. Add decodeUrl, encodeUrl, and decodeUrlIgnoringSpaces functions to it. This wouldn't implement versions of invoke that perform Base64Url encoding/decoding
2. Add a parameter url: Boolean = false to the three functions decode, encode, decodeIgnoringSpaces, plus the two invoke functions. If url=true then perform Base64url encoding/decoding.

These would both provide backward compatibility with existing clients of the Base64 API.

soywiz:
@gsteckman Nice. Go for the one you prefer regarding to the object. For the extension members, can we have the one already exposed, and then for the function ones add a url parameter?
Both would work for me, but I'd say I prefer having as few extension members as possible to not congest the extension scope too much

gsteckman:
I thought of another option. Create a Base64 class and a Base64Url class, similar to what you have with Hash. The encoders will then return Base64/Base64Url based on the encoding option, and the decoders would be overloaded to invoke the proper decoder.

This introduces more classes, but provides a sort of "type safety" on the decoding to make sure a regular Base64 decoding isn't chosen for a Base64url encoded string and vice-versa.

soywiz:
Would that force a change in the API? And for base64, in addition to plain base64 and url encoded one, is there going to be more than two implementations like it happens in Hash? If this is going to be something small and bounded instead of something that could grow beyond two implementations, I would go for something as simple as an "if", if possible, or two separate methods if the other doesn't make sense

gsteckman:
Introducing the new classes would change the API.

There is a Base16 and Base32 defined in the same RFC, but I don't know that they are used much if at all. I've never seen them used before.

I will go with idea #2 above and merge everything into the Base64 object.

krypto/src/commonMain/kotlin/com/soywiz/krypto/encoding/Base64.kt
```diff
@@ -77,15 +102,51 @@ object Base64 {
...
+        }
+    }
+
+    /**
+     * Converts a Base64Url encoded ByteArray to a Base64 encoded ByteArray without adding padding.
+     */
+    private fun base64UrlToBase64(src: ByteArray) {
+        src.forEachIndexed { index, byte ->
+            if (byte == '-'.code.toByte()) {
+                src[index] = '+'.code.toByte()
```
soywiz, line 118: We shouldn't mutate the source array, or that could cause unexpected behaviors. For performance we shouldn't either make a copy, so we should figure out a performant way to handle this while decoding

soywiz:
Nice, Base32 is not typical AFAIK, and Base16 is already in the code base with the Hex name (hexadecimal)

krypto/src/commonMain/kotlin/com/soywiz/krypto/encoding/Base64.kt
gsteckman: I see your point. The prior approach was making a copy. One option is to create a 2nd version of TABLE and DECODE that uses the URL character set, and then use it during encoding/decoding. That would have similar speed performance, with the tradeoff that now the Base64 object has 128 bytes of extra tables. Unless someone is targeting a very low storage/memory embedded system I don't envision that being a problem.

gsteckman: I decided it would be worth the extra 64 bytes to also not create a copy of the encoded data. So implemented a separate TABLE for URL encoding as well.

krypto/src/commonMain/kotlin/com/soywiz/krypto/encoding/Base64.kt
```diff
@@ -8,6 +8,8 @@ object Base64 {
             this[TABLE[n].code] = n
         }
     }
+    private val TABLE_URL = TABLE.toUrlEncode()
+    private val DECODE_URL = DECODE.toUrlDecode()
```
soywiz, line 12: I believe `toUrlEncode` and `toUrlDecode` functions would add more payload size than providing a literal or an inline replacement.

Maybe we can either provide `TABLE_URL` as a literal, or something like `TABLE_URL = TABLE.replace().replace()`.
Since the code for constructing the `DECODE` table is still required, maybe we can extract that to a function (receiving `TABLE` as parameter), and use it for both `DECODE` and `DECODE_URL`.

What do you think?

soywiz: Other than that, I really like what you did here. Storing the right table at the beginning keeps the rest of the code intact, so it is super nice.

gsteckman: I wouldn't expect defining the function as inline vs a separate function to have much impact on the final compiled code size. But I also realized that using `replace()` twice means an extra String array is going to be allocated in memory, and then garbage collected. Plus there is the initialization execution time required when first referencing the Base64 object. We're talking about small array sizes here so it's probably not a big deal, but for the sake of having code uniformity, I made `TABLE_URL` a literal to mirror the `TABLE` literal.

For `DECODE_URL`, there is an advantage to reusing `DECODE`, namely only having to copy it and then modify 2 elements. That should be faster than regenerating a copy directly from the TABLE_URL array. But again the size here is really small so it's probably more important to reduce the amount of code and make it maintainable. So I extracted a function here to be used for both DECODE arrays.

Good suggestions.

soywiz (approved):
LGTM, thanks!, let's wait for CI 👍

soywiz:
Thanks!

base64

PR #425, opened 2022-01-10
Updated buildscripts to add iosSimulatorArm64 target

alexandru-g:
Problem: building for the iOS simulator on an Apple M1 machine would not work.
Solution: added the required iosSimulatorArm64 targets. I've tested using krypto and klock and it works perfectly.

Can someone please take a look if we think we need anything else?

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -91,7 +91,8 @@ val KotlinTarget.isMacosArm64 get() = this.name == "macosArm64"
 val KotlinTarget.isMacos get() = isMacosX64 || isMacosArm64
 val KotlinTarget.isIosArm64 get() = this.name == "iosArm64"
 val KotlinTarget.isIosX64 get() = this.name == "iosX64"
-val KotlinTarget.isIos get() = isIosArm64 || isIosX64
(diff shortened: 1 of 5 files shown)


soywiz:
Uhm. Not sure if I'm missing something, but I believe the ios simulator stuff is already on master:
https://github.com/korlibs/korge-next/blob/b87ad043de5450bad3260646b703871bd028f78c/build.gradle.kts#L95

alexandru-g:
> Uhm. Not sure if I'm missing something, but I believe the ios simulator stuff is already on master:
> 
> https://github.com/korlibs/korge-next/blob/b87ad043de5450bad3260646b703871bd028f78c/build.gradle.kts#L95

This is related to https://github.com/korlibs/korge/issues/620
Are you saying that support for iOS ARM simulators (not x86_64) was already added?

soywiz:
@alexandru-g that's correct. In fact krypto 2.4.10 : https://github.com/korlibs/korge-next/releases/tag/v2.4.10 should already support ios simulator

alexandru-g:
Yup. Just saw it. Perfect then, I'll go ahead and close this!

PR #436, opened 2022-02-08
Bump Kotlin version to 1.6.20-M1

soywiz:


soywiz:
It's broken. Might be related to this: https://blog.jetbrains.com/kotlin/2022/02/kotlin-1-6-20-m1-released/#hierarchical-structure-support-for-multiplatform-projects
That will require carefully checking how it works and adapting the structure to it.

PR #440, opened 2022-02-17
Fix final blit for composed filter

sahilbajaj:


soywiz-invideo:
@sahilbajaj I have fixed master, that had a broken test. Also I have added a new reference test that verifies `Filter` behaviour.
Can you pull from master? Also, since you changed the composed filter, you might need to run `./gradlew jvmTestFix` so the reference test file is updated. That way we will see here the changes in opengl calls for that change.

sahilbajaj:
Done

PR #442, opened 2022-02-18
improve debugger

bognari:
This will hopefully help to make the debug font more readable on height resolutions with small displays.

partial fix / workaround for https://github.com/korlibs/korge/issues/556


korge/src/commonMain/kotlin/com/soywiz/korge/view/Views.kt
```diff
@@ -208,9 +208,10 @@ class Views constructor(
 
     var supportTogglingDebug = true
 	var debugViews = false
-	val debugHandlers = FastArrayList<Views.(RenderContext) -> Unit>()
+	var debugFontExtraScale = 1.0
+	val debugHandlers = FastArrayList<Views.(RenderContext, Double) -> Unit>()
 
-    fun addDebugRenderer(block: Views.(RenderContext) -> Unit) {
+    fun addDebugRenderer(block: Views.(RenderContext, Double) -> Unit) {
```
soywiz, line 214: Instead of passing it as a primitive as an extra argument, can we put this `extraFontScale` inside the `RenderContext`? Either as a property or as an extrinsic Extra.Property? Maybe with a name like `debugExtraFontScale` or something?

Additionally, are you doing this because we are not properly detecting the device pixel ratio on windows? Could we implement that instead?

bognari:
Hey @soywiz I did you requestet change and insertet the debugExtraFontScale  into the RendererContext.

Maybe you are right and it would be a better idea to implement the device pixel ratio correctly but maybe for someone the font is still to small so it is possible to make the font bigger (or smaller).

bognari:
ok now we use the scaling settings too. 
https://stackoverflow.com/questions/32586883/windows-scaling

soywiz:
Awesome! Thanks :) let's wait for CI to finish

korgw/src/jvmMain/kotlin/com/soywiz/korgw/awt/AwtAg.kt
```diff
@@ -15,12 +17,27 @@ class AwtAg(override val nativeComponent: Any, private val checkGl: Boolean, log
...
                 ?: GraphicsEnvironment.getLocalGraphicsEnvironment().defaultScreenDevice.defaultConfiguration
-            return config.defaultTransform.scaleX
+            val scale = config.defaultTransform.scaleX
+
+            // maybe this is not just windows specific :
+            // https://stackoverflow.com/questions/32586883/windows-scaling
+            // somehow this value is not update when you change the scaling in the windows settings while the jvm is running :(
+            val pixelsPerLogicalInchRatio = Toolkit.getDefaultToolkit().screenResolution / defaultPixelsPerLogicalInch
+
+            return@lazy (scale * pixelsPerLogicalInchRatio)
```
soywiz, line 40: Tested on mac. On mac, it should be something like `scale / pixelsPerLogicalInchRatio`.

But I believe we should separate those values:

```
override val devicePixelRatio: Double get() = ... scaleX
override val pixelsPerLogicalInchRatio: Double get() = ... screenResolution / defaultPixelsPerLogicalInch
// Use this in the debug handler, while allowing people to access raw devicePixelRatio without the noise of window scaling
override val computedPixelRatio: Double get() = devicePixelRatio / pixelsPerLogicalInchRatio
```

Otherwise `devicePixelRatio` stops being true. Is it 1 on your hdpi screen? On mac devices with retina displays it is 2, without it, it is 1.

bognari: But yes we can separate the values.

bognari: i'm still not sure about the calculation of "computedPixelRatio" becasue in my opinion it must be 

`devicePixelRatio * pixelsPerLogicalInchRatio` 
or
`pixelsPerLogicalInchRatio / devicePixelRatio `

bognari:
@soywiz any feedback about my pr?

soywiz:
Yes, sorry for the delay. Had a pretty long couple of weeks, and forgot to check this. Im not at home right now, but Ill check it asap

bognari:
sorry. all is good. take your time. this was just a reminder

korge/src/commonMain/kotlin/com/soywiz/korge/render/RenderContext.kt
```diff
@@ -46,6 +46,8 @@ class RenderContext constructor(
     val views: Views? = bp as? Views?
 
     var debugAnnotateView: View? = null
+    var debugExtraFontScale : Double = 1.0
+    var debugExtraFontColor : RGBA = Colors.WHITE
```
soywiz, line 50: Downloaded the PR, and this line seems to fail because RGBA is not found.
You should add a `import com.soywiz.korim.color.*`

soywiz (approved):
Sorry for the delay.

I have tried it and seems to work properly, though you should add a `import com.soywiz.korim.color.*` for it to compile.

Once you add that import, let's merge this ^^

soywiz:
Thanks!

PR #448, opened 2022-03-02
View extension method for scaling view by maintaining aspecg ratio.

Kietyo:


korim/src/commonMain/kotlin/com/soywiz/korim/color/MaterialColors.kt
```diff
@@ -0,0 +1,277 @@
+package com.soywiz.korim.color
+
+// https://material.io/design/color/the-color-system.html#tools-for-picking-colors
+object MaterialColors {
```
soywiz, line 4: Isn't this from another PR?

soywiz (approved):
Can you remove MaterialColors first from the PR?

Kietyo:
> 

Not sure how. Shouldn't be an issue though.

Kietyo:
Ok seems like I fixed it somehow.

soywiz:
Thanks!

PR #450, opened 2022-03-02
Check if View is a Circle when Shape is null

### LINKED ISSUE #449: Circle defaults to BoxShape
If no Shape is specified the registerBodyWithFixture extension function in WorldView defaults to a CircleShape or BoxShape.
It only defaults to CircleShape when the View is an Ellipse and radiusX and radiusY are equal, essentially a circle.
However, because Circle is not a subclass of Ellipse, it defaults to BoxShape, making for odd interactions between objects.

Possible solutions:
1. Refactor Circle ShapeView to be a subclass of Ellipse with equal radiusX and radiusY
2. Remove Circle class and use extension functions to create an Ellipse with equal radiusX and radiusY when a circle is required
3. Add a check if the view is Circle to the when statement in the WordView class

Option 3 will be the least work ;)

As an example, this code:

    circle(50.0, Colors.RED)
      .registerBodyWithFixture(
        type = BodyType.DYNAMIC
      )

Reaches this code in the WorldView class:

    when {
      view is Ellipse && view.isCircle -> CircleShape(view.radiusX / world.customScale)
      else -> BoxShape(getLocalBounds() / world.customScale)
    }

jobhh:
Fixes #449

### DIFF
--- a/korge-box2d/src/commonMain/kotlin/com/soywiz/korge/box2d/WorldView.kt
+++ b/korge-box2d/src/commonMain/kotlin/com/soywiz/korge/box2d/WorldView.kt
@@ -320,6 +320,7 @@ inline fun <T : View> T.registerBodyWithFixture(
     body.fixture {
         this.shape = shape ?:
             when {
+                view is Circle -> CircleShape(view.radius / world.customScale)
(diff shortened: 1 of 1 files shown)


soywiz:
Thanks!

PR #472, opened 2022-03-09
Integrate Fleks ECS into korge + example

jobe-m:
This PR integrates the Entity Component System Fleks as a sub-"module" korge-fleks into the Korge landscape. I have choosen that way because currently it is not possible to add Fleks as an (external) import to the example code.

The Fleks repo (https://github.com/Quillraven/Fleks) contains a branch "kmp" which in principle contains the same code as under "korge-fleks" in this PR. The kmp branch was pushed by me into Fleks. It contains all necessary changes to get the Fleks codebase usable with Korge (and Kotlin multiplatform in general - I guess). But Fleks is currently exported in a "multiplatform" incompatible way to maven and therefore it can not yet be used in a Korge project.

The Fleks repo and also "Korge-fleks" contains unit tests which were also adapted to the changes for mulitplatform.

The added example "fleks-ecs" shows the ECS system in action :) It creates a lot of objects which are drawn out from one sprite animation. I hope that it shows how to use Fleks in a game. When I have time I will update the example sources with more comments and descriptions. Maybe we can also create a section for Fleks on the Korge documentation pages.

~~Currently I only have one problem. It looks like that the sources from korge-fleks are not exported properly to maven-local. Thus I cannot use Fleks in my own game project. Maybe there is some configuration in gradle missing which would properly export Fleks from Korge-fleks into maven so that it can be used in stand alone projects by just adding Korge as a dependency.~~ This is fixed.

Also I don't know if you want to keep Fleks as part of Korge for long time or if you would like to have it externalized (in its own origin git repo from Quillraven). The benefit of having Fleks in Korge would be that we could adapt, optimize and integrate it further to be a companion to Korge. E.g. writing Systems which are integrated into Korges animation/image views, etc. So that anybody who wants to use the ECS with Korge do not need to write an "Animation" or "Sprite system" by its own.


soywiz:
😲 Awesome!

I'll check in detail later this week, since this is pretty big! In the mean time, CI seems to be broken somehow: 
https://github.com/korlibs/korge-next/runs/5481838669?check_suite_focus=true#step:6:228

Maybe it is using junit 5, while the korlibs are using junit 4?

Quillraven:
@soywiz I am using Kotlin test 1.6.0 dependency with the useJUnitPlatform configuration as seen in the build.gradle file: https://github.com/Quillraven/Fleks/blob/master/build.gradle.kts

I personally have no preference about which testing framework is used. So it can be changed to whatever is compatible with KorGE but do you know which adjustments need to be done in the Gradle file?

Also, I have almost no experience with KMP and I think @jobe-m has little Gradle experience. Do you know someone who could help with setting up the build file properly and the sourcesets to correctly publish it to maven? 

I already commented it in the KMP issue of the repository but at the moment I have little time to get more familiar with the KMP specifics :(

jobe-m:
Thanks for already taking a look at my PR guys :)
When it is agreed that we keep fleks for now in Korge I would look into the unit tests if they need changes to make them passing also in the Korge environment.
@soywiz please take your time, no hurry :)

jobe-m:
I fixed some issues in the module tests. But it looks that there are still exceptions when running the tests for Kotlin native. I guess the problem is on the code side. Not in the unit tests. Because when I try to compile and run the fleks example on Kotlin native than running the kexe crashes.
I will take a look and try to find the problem.

soywiz:
@jobe-m for InvalidMutability exception, you have to mark global variables with `@ThreadLocal`

korge-fleks/src/commonMain/kotlin/com/github/quillraven/fleks/collection/bitArray.kt
```diff
@@ -0,0 +1,156 @@
...
+            hash = 127 * hash + (bits[i] xor (bits[i] ushr 32)).toInt()
+        }
+        return hash
+    }
+
+    override fun equals(other: Any?): Boolean {
+        if (other == null) return false
+        if (this === other) return true
+
+        other as BitArray
```
soywiz, line 146: This will fail instead of returning false, in the case `other` is from another type, right?

soywiz (approved):
LGTM. Really nice!

BTW. Is the PR ready to merge, or something is missing? (it is still in draft)

Could you move the `ImageAnimation.Direction.ONCE_*` functionality to another PR?
So in the next release appears as new functionality not tied only to fleks?

korge-fleks/src/commonMain/kotlin/com/github/quillraven/fleks/collection/bitArray.kt
jobe-m: I will check :)

jobe-m:
> Could you move the `ImageAnimation.Direction.ONCE_*` functionality to another PR? So in the next release appears as new functionality not tied only to fleks?

Yes, good idea. I will do it, too.

jobe-m:
YES :)
I got korge-fleks properly exported so that it can be used by an external project which includes Korge.
For that I needed to add a "supportFleks()" function into the korge-gradle-plugin.

korge-fleks/src/commonMain/kotlin/com/github/quillraven/fleks/collection/bitArray.kt
```diff
@@ -140,10 +141,11 @@ class BitArray(
     }
 
     override fun equals(other: Any?): Boolean {
-        if (other == null) return false
         if (this === other) return true
+        if (other !is BitArray) throw FleksBitArrayTypeException(
```
soywiz, line 145: Equals shouldn't throw: https://docs.oracle.com/javase/7/docs/api/java/lang/Object.html#equals(java.lang.Object)

instead it should return false.
Suppose for example you have a set of mixed objects, using a HashSet and one object of a different type collides its hashCode. It the implementation will check object.equals() with objects with the same hashCode but different type. And it will expect to be false for different instances, and to not throw

soywiz: So `if (other !is BitArray) return false`

jobe-m: You are right. I will change it :)

jobe-m:
@soywiz the PR should be now ready to merge :)

soywiz (approved):
LGTM! 🚀

jobe-m:
Wow, finally it's ready and usable from Korge. Thanks to @Quillraven for creating Fleks and sharing it with us all :)

soywiz:
Yeah +1. Thanks @Quillraven & thanks @jobe-m !

fleks

Quillraven:
That's really cool, thanks guys!

@jobe-m can you please create another PR for Fleks with the important changes? I think I will make a "next" branch for the next version that combines KMP and the JVM version. We can put your changes then already in this branch for now.

PR #488, opened 2022-03-14
Implement playing image animation only once

jobe-m:
This commit implements playing an image animation only one forward or
reverse. For that it is possible to set the direction to ONCE_FORWARD or
ONCE_REVERSE. If direction is "ONCE_*" then when playing has finished
the call-back onPlayFinished() will be invoked. With that is is possible
to react on finished animation playing like scrap the animationView
object or reset it to be used again.

Also there was a second call-back added: onDestroyLayer()
That is invoked when the layers of the animation view gets deleted
because the animation is changed to another in ImageData. That call-back
can be used to recycle the image objects which are used for the image
animation view layers.


soywiz (approved):
Nice thanks! 🎉

PR #507, opened 2022-03-20
Delete samples/box2dv2 directory

SchneckeddeF:
theres nothing in this directory and the last update of it was in 2020


soywiz:
thanks!

PR #536, opened 2022-03-25
Fix loading frame cells in Aseprite files

jobe-m:
Then the Aseprite file contains empty cells in a to-be-loaded layer then
it did not create ImageFrame objects for those empty cells. That lead to
the problem that an animation could not be played if there are empty
cells in frames before the animation tag.
Now empty ImageFrame objects are created for empty cells. This fixes the
problem.

Unit test added to check if the frame array in the loaded Aseprite image
has the correct length.


soywiz:
Nice. Thanks!

PR #537, opened 2022-03-25
Update Fleks ECS

jobe-m:
New feature was added which makes it possible to mark an injectable as
"used". Thus Fleks will not throw exception if the injectable is not
used directly at creation of system and listener objects. That means it
is now possible to use the injected objects during init functions. Default
for "used" option is "false". Which has the same effect as before.


soywiz (approved):
LGTM 👍

PR #561, opened 2022-04-02
Antialiased GPU rendering

soywiz-invideo:
Some work on https://github.com/korlibs/korge/issues/550

This includes stroke AA rendering with [mittered](https://www.dacostabalboa.com/es/imagenes/html5_canvas.png) joins and butt caps


soywiz-invideo:
TODO: we have to shrink shapes a bit to draw the AA stroke

PR #589, opened 2022-04-15
Handle konan cache separately

soywiz:


soywiz:
Doesn't seems to improve reasonable

PR #592, opened 2022-04-15
Have Container implement MutableCollection

Kietyo:
Provides a ton of useful extension functions from the standard library.


soywiz (approved):
LGTM 👍
I guess it is time to reformat all the code, and use a linter here :)

PR #615, opened 2022-05-02
Utils for distributing views evenly across vertically/horizontally.

Kietyo:
Real life example:

```
	val numbers = container {
		repeat(5) {
			val num = it + 1
			UIButton(BUTTON_WIDTH, BUTTON_HEIGHT, num.toString()).apply {
				onClick {
					if (checkBox.checked) {
						sudokuBoard.applyHint(num)
					} else {
						sudokuBoard.applyNumber(num)
					}
				}
			}.addTo(this)
		}
	}
	distributeEvenlyHorizontally(numbers.children, sudokuBoard.scaledWidth)
```

image

### DIFF
--- a/korge/src/commonMain/kotlin/com/soywiz/korge/view/util/util.kt
+++ b/korge/src/commonMain/kotlin/com/soywiz/korge/view/util/util.kt
@@ -1,0 +1,106 @@
+package com.soywiz.korge.view.util
+
+import com.soywiz.korge.view.View
+
+
+// Distributes an iterable of views evenly across a provided bounding width.
+// The first view in the iterable will act as the ANCHOR VIEW which all other views will be
+// offset from.
+//
+// If you've used Google Slides/Microsoft Excel before, this is similar to evenly distributing a
+// bunch of selected shapes HORIZONTALLY.
+//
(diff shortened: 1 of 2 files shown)


soywiz:
Nice!

BTW. Are you aware of the `uiVerticalStack { }`,`uiHorizontalStack { }`,`uiHorizontalVerticalFill { }` etc.?
Before merging I would like to understand if you have tried them and if those would work for you?

Example: https://github.com/korlibs/korge-next/blob/fd5dd1525eacdbf81549257732ffe94e63ad046a/korge-sandbox/src/commonMain/kotlin/MainBlur.kt#L74
You can uncomment this line and execute `./gradlew :korge-sandbox:runJvm`: https://github.com/korlibs/korge-next/blob/fd5dd1525eacdbf81549257732ffe94e63ad046a/korge-sandbox/src/commonMain/kotlin/Main.kt#L40

Kietyo:
I've tried them and my biggest pet peeve with them is that they modify the child views themselves, changing their widths and height.

uiHorizontalFill:

image

uiHorizontalStack:

image

Note that the utils I introduce do not change the views width or height. It is very similar to this functionality here in Google Slides:

image

soywiz:
Perfect.

`T H A N K S !`

:)

PR #616, opened 2022-05-02
Have *Array2 classes implement an interface [Includes small breaking change]

Kietyo:
This will allow me to add more useful extension functions in the future without have to redundantly implement it for each type.

**There is a small breaking change.**

I changed `toString(...)` method names to `asString` since I found refactoring it broke the JS IR compiler. I don't think this will be that big of change because the no-parameter `toString()` implementation is still the same as before, and that's probably what will be used the most. 

Clients can simply rename toString -> asString if they're using the overloaded versions.


kds/src/commonMain/kotlin/com/soywiz/kds/Array2.kt
```diff
@@ -241,14 +290,12 @@ data class IntArray2(val width: Int, val height: Int, val data: IntArray) : Iter
...
-        var n = 0
-        for (y in rows.indices) {
-            val row = rows[y]
-            for (x in row.indices) {
-                this.data[n++] = row[x]
-            }
-        }
+    override fun set(idx: Int, value: Int) {
+        this.data[idx] = value
+    }
```
soywiz, line 295: This would incur in boxing, and that would break the whole point of having an specialized class. Please check: https://github.com/korlibs/korge-next/commit/86ec6e0cc08d629d2a92aeaf23c849e776f44cad#diff-528f00b875719471f92b82e3e23e9d79bda46d3b03427de315cd6a863f4fa261L60-L61

So what I would suggest is to have methods like `getAt` and `setAt` that are generic, but the specialized one is not in the interface, to prevent a regression in performance.

What do you think=

soywiz:
Nice. But we should be aware of the performance regression due to boxing.

Also what's the JS bug? If we circumvent it, we should also probably report it. What's the stacktrace?

Kietyo:
Thanks a lot for the comments, I'm learning a lot! I also saw your post a few years back:


I decided to go with revert the get/set changes and introduce the getAt and setAt methods that you mention.

Next I'm working on a potential workaround to avoid extra allocs for the general methods (e.g each, fill), which is to introduce an auxiliary data class that can be used to passing data around without boxing:

```
data class Array2Element(
    var intData: Int = 0,
    var floatData: Float = 0f,
    var doubleData: Double = 0.0,
    var any: Any? = null
)
```

I took a look at the generated Java code, and noticed it doesn't box:

image

I'll work more on it tomorrow.

kds/src/commonMain/kotlin/com/soywiz/kds/Array2.kt
Kietyo: SGTM

Kietyo:
Ok NVM the idea with using an auxiliary data didn't work. Ready for review.

soywiz:
Thanks!

PR #618, opened 2022-05-02
the mapping for cyan and magenta was missing

bognari:
the mapping for cyan and magenta was missing so it was not usable with Colors["CYAN"] and Colors["magenta"]


soywiz:
imageonline-co-textimage (2)

PR #625, opened 2022-05-06
Implement closeable mouse events & closeable draggable.

Kietyo:
My use case for this is I wanted to be able to disable draggable.

For example in my game, there is an **editor mode** where users can place rocks on a map. When the user is in this mode, I want to disable dragging.


korge-sandbox/src/commonMain/kotlin/MainDraggable.kt
```diff
@@ -0,0 +1,22 @@
...
+
+suspend fun Stage.mainDraggable() {
+    val draggableRect = solidRect(100, 100)
+
+    var closeable: DraggableCloseable? = null
+
+    uiCheckBox(text = "Toggle dragging") {
+        onChange {
+            if (it.checked) {
+                closeable = draggableRect.draggableCloseable().second
```
soywiz, line 16: Since we have the view already, what about just returning the closeable? draggble was returning the instance to do chaining, but you cannot do chaining anyway if you need the closeable instance too

Kietyo: Maybe we can just default to returning Closeable anyways, so we don't need to split to two methods? I dont think chaining is needed since Kotlin has scoping functions. WDYT?

soywiz: That would be a breaking change. Id suggest to keep the two methods for now, and maybe we can deprecate the original one. The other thing is that most of the methods extending views are chainable so to be consistent the API is like that.
Despite Kotlin is offering scope functions, one-liners still benefit from chaining. For example:
`val rect = solidRect(100, 100, Colors.RED).xy(200, 300).centered`

Kietyo: Done

soywiz:
Thanks!

PR #642, opened 2022-05-11
Atlas with rotated textures

### LINKED ISSUE #622: Atlas with rotated textures
I'm trying to parse a given texture map into an Atlas (Sparrow/Starling format).

But I can't get the desired result. As you can see in the texture map image, the textures are partially rotated.

If I don't do applyRotation() on the AtlasInfo.Region, the texture is rotated but the aspect ratio is wrong. (See result picture first row)

When I do the applyRotation() the aspect ratio is correct but the texture coordinates are wrong. (See result image second row)

Am I doing something wrong here, or is this a bug?

Result:
atlas_test

Texture map image:
test

Texture map atlas:
```xml
<?xml version="1.0" encoding="UTF-8"?>
<!-- Created with TexturePacker https://www.codeandweb.com/texturepacker -->
<!-- $TexturePacker:SmartUpdate:8a19fb2bb2eb30d8c28d33462b2198ec:d7a70b121343375e5ed7a6e02819199e:25a2456976bfe59cdd62c104b7067021$ -->
<TextureAtlas imagePath="test.png" width="128" height="128">
    <SubTexture name="1" x="37" y="73" rotated="true" width="47" height="33" frameX="0" frameY="0" frameWidth="42" frameHeight="56"/>
    <SubTexture name="2" x="1" y="1" width="34" height="48" frameX="-4" frameY="-4" frameWidth="42" frameHeight="56"/>
    <SubTexture name="3" x="1" y="51" width="34" height="48" frameX="-4" frameY="-4" frameWidth="42" frameHeight="56"/>
    <SubTexture name="4" x="37" y="1" rotated="true" width="48" height="34" frameX="-4" frameY="-4" frameWidth="42" frameHeight="56"/>
    <SubTexture name="5" x="87" y="1" width="34" height="48" frameX="-4" frameY="-4" frameWidth="42" frameHeight="56"/>
    <SubTexture name="6" x="37" y="37" rotated="true" width="48" height="34" frameX="-4" frameY="-4" frameWidth="42" frameHeight="56"/>
    <SubTexture name="7" x="87" y="51" width="34" height="48" frameX="-4" frameY="-4" frameWidth="42" frameHeight="56"/>
</TextureAtlas>
```

Code:
```kotlin
suspend fun main() = Korge(width = 256, height = 256, virtualWidth = 256, virtualHeight = 256, bgcolor = Colors["#2b2b2b"]) {
	var atlas = parseAtlas("test.xml")
	var x = 0.0
	for (i in 1..7) {
		image(atlas["$i"]).xy(x, 0.0).also {
			x += it.width
		}
	}

	atlas = parseAtlas("test.xml", true)
	x = 0.0
	for (i in 1..7) {
		image(atlas["$i"]).xy(x, 100.0).also {
			x += it.width
		}
	}
}

suspend fun parseAtlas(atlasFile: String, applyRotation: Boolean = false): Atlas {
	val xmlContent = resourcesVfs[atlasFile].readString(Charsets.UTF8)
	val xml = Xml(xmlContent)
	val imagePath = xml.str("imagePath")

	val atlasImg = resourcesVfs[imagePath].readBitmap().sliceWithSize(0, 0, xml.int("width"), xml.int("height"))
	val atlasRegions = mutableListOf<AtlasInfo.Region>()

	xml.children("SubTexture").forEach {
		val name = it.str("name", "")
		if (name.isEmpty()) {
			return@forEach
		}
		val rotated = it.boolean("rotated", false)
		val texFrame = AtlasInfo.Rect(
			it.int("x"),
			it.int("y"),
			it.int("width"),
			it.int("height")
		)
		val trimmed = it.hasAttribute("frameX")
		val imageFrame =
			if (trimmed) AtlasInfo.Rect(
				it.int("frameX") * -1,
				it.int("frameY") * -1,
				it.int("frameWidth"),
				it.int("frameHeight")
			)
			else texFrame

		val region = AtlasInfo.Region(
			name = name,
			frame = texFrame,
			rotated = rotated,
			sourceSize = AtlasInfo.Size(imageFrame.w, imageFrame.h),
			spriteSourceSize = imageFrame,
			orig = AtlasInfo.Size(imageFrame.w, imageFrame.h),
			trimmed = trimmed
		)

		atlasRegions += if (applyRotation) region.applyRotation() else region
	}

	val atlasInfo = AtlasInfo(
		frames = atlasRegions,
		meta = AtlasInfo.Meta(
			app = "Sparrow/Starling",
			format = "RGBA8888",
			image = imagePath,
			scale = 1.0,
			size = AtlasInfo.Size(xml.int("width"), xml.int("height")),
			version = AtlasInfo.Meta.VERSION
		)
	)
	return Atlas(atlasImg, atlasInfo)
}
```

Stechmuck:
This pull request adds an optional parameter `textureCoords` to `AtlasInfo.Region` and `BmpSlice`.

If `textureCoords` is specified, the values ​​for the `BmpCoords` interface in `BmpSlice` are no longer determined from the bounds variable, but taken directly from `textureBounds`.

This resolves #622

**Background:**

The file formats txt, xml and json define the frame for a rotated texture in the atlas in different ways

* txt (spine) rotates the texture counter-clockwise and defines the frame without rotation.
* xml (Sparrow/Starling) rotates the texture clockwise and defines the frame with rotation.
* json (spriter) rotates the texture clockwise and defines the frame without rotation.

As a result, the current implementation does not work for all three formats.
With the solution implemented here, all three formats are now read in such a way that with a rotated texture the bounds in `BmpSlice` are always unrotated and `rotated` is set to false.

In addition, I added rotation and virtualFrame support when reading an xml file, and virtualFrame support for a txt file

**Result:**

```kotlin
suspend fun main() = Korge(width = 256, height = 256, virtualWidth = 256, virtualHeight = 256, bgcolor = Colors["#2b2b2b"]) {

	fun generateImages(offsetY: Double, atlas: Atlas, ext: String = "") {
		var x = 0.0
		for (i in 1..7) {
			image(atlas["$i$ext"]).xy(x, offsetY).also {
				x += it.width
			}
		}
	}

	generateImages(0.0, resourcesVfs["test.xml"].readAtlas())

	generateImages(80.0, resourcesVfs["test2.atlas.txt"].readAtlas())

	generateImages(160.0, resourcesVfs["test.json"].readAtlas(), ".png")
}
```
testresult


Stechmuck:
Because of the failing test I've created a class `TextureCoords:BmpCoords` and use this class instead of the object:BmpCoords. All references are renamed

Stechmuck:
Unfortunately, they don't really help, since the rotation must already be known when the AtlasInfo.Region is created and this is not an instance of BmpCoordsWithT<T>.

If at all you can use these functions when creating the BmpSlice. For this, however, it would have to be stored in the AtlasInfo.Region how the BmpSlice has to be rotated.
CW? CCW? Rotate by how many degrees? Where is the anchor? Or maybe a completely different transformation?
Who knows what other formats are out there.

soywiz:
I see. My concern is that looking at the code, there are several new ifs and arithmetic regarding to computing coordinates, and there are a couple of computations that look similar, and might be generalized or they might be available already, no new tests covering those cases, and one of the tests changes the `firstFrame.rotated` from false to true (I guess that it is because now they are precomputed, so no need for rotation to be marked).
And since there was functionality for that I was wondering if we could somehow reuse that.
I believe when the code was created we didn't have texture coordinates yet, and that's why the rotated was there to be taken into account in the very first moment. But indeed a rotation boolean is a bit lacky in terms of what does it mean.

In latests versions we introduced https://github.com/korlibs/korge-next/blob/cb355a04c6fc530ca63eed71987a101f6feb6660/korim/src/commonMain/kotlin/com/soywiz/korim/format/ImageOrientation.kt 
and it can be applied with this:
https://github.com/korlibs/korge-next/blob/cb355a04c6fc530ca63eed71987a101f6feb6660/korim/src/commonMain/kotlin/com/soywiz/korim/format/ImageOrientation.kt#L40
so that might help to better describe the kind of transformation we need to compute it at that point or later, if the flippedX, flippedY and rotatedLeft, rotatedRight doesn't fit.

The `BmpCoordsWithT<T>` is generic so it can support all kind of bases and not only Bitmaps.

You can either use:
https://github.com/korlibs/korge-next/blob/cb355a04c6fc530ca63eed71987a101f6feb6660/korim/src/commonMain/kotlin/com/soywiz/korim/bitmap/BitmapSlice.kt#L78

or:
https://github.com/korlibs/korge-next/blob/cb355a04c6fc530ca63eed71987a101f6feb6660/korim/src/commonMain/kotlin/com/soywiz/korim/bitmap/BitmapSlice.kt#L96

Since the generic should be a `ISizeInt` you can use as base a plain `SizeInt` with the width and height of the full image I guess.

So instead of creating the `TextureCoords` class I'd suggest you to use `BmpCoordsWithInstance<ISizeInt>` and use `rotatedLeft` etc.
And if you miss some functionality in those clasess, maybe we can add them so it is more genereric and the atlas functionality is simpler without branches or complicated edge cases.
What do you think?

Stechmuck:
OK, I didn't know about ImageOrentation.

Knowing this, I had the idea of equipping AtlasInfo.Region with an ImageOrientation instead of the TextureCoords.

This should make it possible to apply the ImageOrientation to the BitmapSlice when creating it in Atlas.Entry.slice

Does that sound reasonable?

If that's a solution, what do you think of the idea of changing the primary constructor of AtlasInfo.Region to

```kotlin
     data class Region(
         val name: String,
         val frame: Rect,
         val virtFrame: Rect? = null,
         val imageOrientation: ImageOrientation = ImageOrientation.ORIGINAL
     )
```

and then implement the current constructor as a secondary constructor, mark it as `@Deprecated` along with the missing variables (implemented as properties)?

Then the whole thing should be backwards compatible and would be a bit easier to handle in the future

Stechmuck:
@soywiz 

I now have a version of BmpSlice with which it is possible to extract rotated slices using ImageOrientaion. However, now one of 2 tests always fails. Depending on how I treat the offset in a slice from a slice.

In [NinePatchBitmap32Test.name()](https://github.com/korlibs/korge-next/blob/2eb8bb02f18cfa09b83baf62669766302b4a16bc/korim/src/commonTest/kotlin/com/soywiz/korim/bitmap/NinePatchBitmap32Test.kt#L35) a slice `x:1 y:1 w:200 h:200` is extracted from a `202x202` bitmap (`ninePatch.info.content`). For example, another slice with `x:0 y:0 w:90 h:56` is then created from this slice and the offset (x,y) is expected to be `0,0`.

In [BitmapSliceTest.test()](https://github.com/korlibs/korge-next/blob/2eb8bb02f18cfa09b83baf62669766302b4a16bc/korim/src/commonTest/kotlin/com/soywiz/korim/bitmap/BitmapSliceTest.kt#L8) a slice `x:16 y:16 w:32 h:32` is extracted from a `64x64` bitmap. From this slice, another slice is then created with `x:8 y:8 w:40 h:40` and expects the offset to be `24,24`.

So in the first test `1,1 + 0,0 = 0,0` (offset within the parent slice) and in the second test `16,16 + 8,8 = 24,24` (offset within the base bitmap)

---

Most functions in BmpSlice so far are not really able to handle transformation (`ImageOrientation`) or the `virtFrame`.

##### `sliceXXX()` with `virtFrame`

* Should these functions return a slice with a corresponding `virtFrame`?
  For example `bounds[x:20 y:20 w:20 h:20] virtFrame[x:20 y:20 w:80 h:80]` -> `sliceWithSize(10, 10, 20, 20)` -> ` bounds[x:0 y:0 w:10 h:10] virtFrame[x:10 y:10 w:20 h:20]`

##### `get/setRgba()`, `readPixels()` and `extract()`

* These functions are currently based on the `bounds` of `BmpSlice`. With transformed slices you would have to switch to the `BmpCoords`, which complicates the calculation of the coordinates for each pixel.

* Should the `virtFrame` also be taken into account here? And then what to do if `setRgba()` is on a point outside the `BmpCoords` but inside the `virtFrame`?

##### `flippedX/Y()`, `rotatedLeft/Right()` and `splitInRows()`

* Should the `virtFrame` also be transformed here?

soywiz:
@Stechmuck what do you think it would be the best approach? And/or the simplest one?

Stechmuck:
@soywiz 

From a user's point of view, I would say that `NinePatchBitmap32Test.name()` expects the right thing. Because why should the offset I specified change?

So I would change `BitmapSliceTest.test()` so that it expects an offset of `8,8`.

---

Also from the user's point of view... I only know the original image. I don't care if it was trimmed or rotated in an atlas. I don't care about `virtFrame` or `BmpCoords`. As such, I would expect all functions to respect `virtFrame`.

In the special case of `setRgba()` I would (if necessary) use `extract()` to extract a bitmap and set it as a new `base` in which I can then set the color of the pixel.

I think that would then require defining `base` as a `var` with a `private set`

soywiz:
@Stechmuck sorry forgot to reply. Then if you believe that won't cause strange behaviours and makes sense, feel free to do that. It is okay to change tests assets if it makes sense 👍

Stechmuck:
@soywiz

I have now made the following changes:

`Atlas.Region`
* The old constructor is now deprecated and replaced with `(name, frame, virtFrame, imageOrientation)`
* xml supports `virtFrame` and rotation

Moved `virtFrame` to `BmpCoordsWithT<T : ISizeInt>`.

`BmpCoordsWithT<T>.rotatedLeft() / .rotatedRight() / .flippedX() / .flippedY()` support `virtFrame`

`BmpCoordsWithT<T>.transformed(m: Matrix)` doesn't support a `virtFrame` because I don't know what it should look like then. Also, `transformed()` doesn't work in `Atlas` because `bmpCoords` appear to form new `bounds` around the transformed slice, thus including neighboring images.

`BmpSlice`
* Constructor with `rotated` is depracated and replaced with `imageOrientation` instead of `rotated`
* now has an `init{}` block that makes it possible to create further (transformed) slices from transformed slices (including `virtFrame`)
* `read/writePixel()` / `get/setRgba()` respect the `virtFrame`
* if `setRgba()` sets a pixel outside the `BmpCoords` but inside the `virtFrame`, the function extracts a new bitmap and sets it as the new base. To do this, I had to change a few values ​​in properties with `private set()`.

`BmpSlice.extract()` and `BitmapSlice.extract()` respect `virtFrame` and transformation. In other words, the created bitmap has the size `frameWidth x frameHeight`, an offset of `frameOffsetX, frameOffsetY` and no more transformation.

Added tests for new `BitmapSlice` functions.
Added tests for `Atlas` `xml`, `txt` and `json` with new slice functions.

The test `BitmapSliceTest.test()` has been modified so that the last slice now expects an offset of `8,8`

soywiz:
I will try to check during the weekend. Thanks for the work!

korim/src/commonMain/kotlin/com/soywiz/korim/bitmap/BitmapSlice.kt
```diff
@@ -281,68 +329,313 @@ abstract class BmpSlice(
...
+
+    fun getRgba(x: Int, y: Int): RGBA {
+        basePixelPos(x, y)?.let {
+            return bmpBase.getRgba(it.x, it.y)
+        }
+        return Colors.TRANSPARENT_BLACK
+    }
+
+    fun setRgba(x: Int, y: Int, value: RGBA) {
+        basePixelPos(x, y).also {
```
soywiz, line 544: This seems to be allocating, maybe not acceptable here because this kind of method is going to be executed like a lot of times.

Not sure about the details but what about creating two methods:

basePixelPosX and basePixelPosY, so it can return a simple Int and no allocation is needed? You can still keep basePixelPos calling both of them to keep the method is external consumers can accept that trade-off

korim/src/commonMain/kotlin/com/soywiz/korim/bitmap/BitmapSlice.kt
```diff
@@ -281,68 +329,313 @@ abstract class BmpSlice(
...
+                bounds.setBoundsTo(0, 0, bmpBase.width, bmpBase.height)
+                virtFrame?.setBoundsTo(0, 0, bmpBase.width, bmpBase.height)
+                bmpCoords = BmpCoordsWithInstanceBase(
+                    SizeInt(bmpBase.width, bmpBase.height),
+                    0f, 0f, 1f, 0f, 1f, 1f, 0f, 1f
+                )
+                intArrayOf(0, 0, 1, 0, 1, 0).copyInto(pixelOffsets)
+                imageOrientation = ImageOrientation.ORIGINAL
+                bmpBase.setRgba(x, y, value)
+            }
```
soywiz, line 561: What's all this code about? If `basePixelPos` is null, shouldn't this just do nothing since it is outside?

korim/src/commonMain/kotlin/com/soywiz/korim/bitmap/BitmapSlice.kt
```diff
@@ -281,68 +329,313 @@ abstract class BmpSlice(
...
+    override val bl_y: Float get() = this.bmpCoords.bl_y
 
+    @Deprecated("Use imageOrientation")
+    val rotated: Boolean = imageOrientation == ImageOrientation.ROTATE_90 || imageOrientation == ImageOrientation.ROTATE_270
     val rotatedAngle: Int = 0
 
+    private val pixelOffsets: IntArray by lazy {
+        val x = (tl_x * baseWidth).roundToInt()
+        val y = (tl_y * baseHeight).roundToInt()
+        if (tl_x == tr_x) {
```
soywiz, line 485: I have seen this comparison in a couple of places. What's the meaning of that comparison? When that is true how is the BmpSlice?

Can we have something like:

```kotlin
class BmpSlice {
  val isSomething: Boolean get() = tl_x == tr_x

  ...
  if (isSomething) {
    ...
  }
}
```

korim/src/commonMain/kotlin/com/soywiz/korim/bitmap/BitmapSlice.kt
```diff
@@ -281,68 +329,313 @@ abstract class BmpSlice(
...
+                    when (imageOrientation.rotation) {
+                        ImageOrientation.Rotation.R90, ImageOrientation.Rotation.R270 -> {
+                            brX = tlX + w / bh * dx
+                            brY = tlY + h / bw * dy
+                        }
+                        else -> {
+                            brX = tlX + h / bh * dx
+                            brY = tlY + w / bw * dy
+                        }
+                    }
```
soywiz, line 382: Looking at this, can we simplify this kind of things?

For example:

```kotlin
val rotated = imageOrientation.rotation == ImageOrientation.Rotation.R90 || imageOrientation.rotation == ImageOrientation.Rotation.R270
val xDim = if (rotated) w else h
val yDim = if (rotated) w else h
brX = tlX + xDim / bh * dx
brY = tlY + yDim / bw * dy
```

The idea is to reduce duplicated arithmetic and ifs as much as possible so the code is as straightforward as possible. Since there are many big ifs and stuff there is a lot of code and it is hard to view all the edge cases, and changing something might lead to not changing all the paths, etc.
Can you try to figure out patterns in the code and simplify this code a bit more?

korim/src/commonMain/kotlin/com/soywiz/korim/bitmap/BitmapSlice.kt
Stechmuck: `null` means, it is outsite the `bmpCoords`. In the else there is a check if it is inside the `virtFrame` and if so the slice will be extracted and the pixel is set in the new extracted Bitmap.

That is the Point
> * if `setRgba()` sets a pixel outside the `BmpCoords` but inside the `virtFrame`, the function extracts a new bitmap and sets it as the new base. To do this, I had to change a few values ​​in properties with `private set()`.

korim/src/commonMain/kotlin/com/soywiz/korim/bitmap/BitmapSlice.kt
Stechmuck: That is, if the `BmpSlice` is rotated 90° CW or CCW relative to the Bitmap. Then `tl_x` is equal to `tr_x` and `width` and `height` must be flipped in calculations between `bounds` and `bmpCoords`. (That is not `true` for example for a slice of a slice where both are rotatet 90°)

I suggest this name and the comparsion should be also changed to be correct.
``` kotlin
val isRotatedInBmpDeg90 = (width > 1 && tl_x == tr_x) || (height > 1 && tl_y == bl_y)
```

korim/src/commonMain/kotlin/com/soywiz/korim/bitmap/BitmapSlice.kt
Stechmuck: Found a little problem,..

To calculate the `basePixelPos` I need to know the `x` and `y` (because of the "`if (tl_x == tr_x) {` problem"). So if I have `basePixelPosX` and `basePixelPosY` most of the validation and calculation have to be done twice.

But `basePixelPos` has an optional parameter `out: PointInt = PointInt()`. Do you think it's ok to create a `private tmpPointInt = PointInt()` and let `set/getRgba()` use it / have an optional parameter `out: PointInt = tmpPointInt`, or will this lead to multithreading problems?

korim/src/commonMain/kotlin/com/soywiz/korim/bitmap/BitmapSlice.kt
```diff
@@ -250,29 +294,33 @@ fun <T : ISizeInt> BmpCoordsWithT<T>.named(name: String?): BmpCoordsWithInstance
...
-    private val offset = if (rotated) 1 else 0
+    var bmpBase: Bitmap = bmpBase
+        private set
+    var imageOrientation: ImageOrientation = imageOrientation
+        private set
 
-    private val p0 = points.getCyclic(offset + 0)
-    private val p1 = points.getCyclic(offset + 1)
-    private val p2 = points.getCyclic(offset + 2)
-    private val p3 = points.getCyclic(offset + 3)
```
soywiz, line 275: Can't we use these points somehow or an array of points and represent the `ImageOrientation` as a reorder of those points?

Like flippedX, flippedY:

https://github.com/korlibs/korge-next/blob/44300cc63f05d20ff25bf94a077bbb47b29ee704/korim/src/commonMain/kotlin/com/soywiz/korim/bitmap/BitmapSlice.kt#L215-L225

The `init { }` method in this PR looks pretty complicated to me and doing a lot of things with lots of edge cases and I'm a bit afraid of introducing it and having to maintain it later.

I'm also wondering if we really need to put this complexity here. In fact those methods are already doing the required transformations.
What if instead of returning `BmpSlice` in the atlas, we return values from those methods, and we apply the required transformations there?

BmpCoordsWithT was introduced precisely for that, to keep BmpSlice simple, and maybe we could even deprecate the rotated property and ignore it. What would you think? Is something preventing from returning that in the atlas? Since we are bumping version to 3.0.0 and it is a major release, we can afford to change return types

In fact in KorGE views we alreaady changed:
https://github.com/korlibs/korge-next/blob/44300cc63f05d20ff25bf94a077bbb47b29ee704/korim/src/commonMain/kotlin/com/soywiz/korim/bitmap/BitmapSlice.kt#L84

To use BaseBmpSlice:
https://github.com/korlibs/korge-next/blob/44300cc63f05d20ff25bf94a077bbb47b29ee704/korge/src/commonMain/kotlin/com/soywiz/korge/view/Image.kt#L46

That is the type returned from those methods.

Stechmuck:
The points `p0` to `p3` as a representation for the `ImageOrientation` is nothing other than the `BmpCoords`, isn't it?

The `init { }` function is mainly there to calculate the correct `BmpCoords` based on the parent element. If the `BmpCoords` of the parent element are rotated 90° it gets a bit complicated.

The reason I moved this to `BmpSlice` is that a user might want to create another `BmpSlice` from a `BmpSlice` that they got from an `Atlas` for example. If the `BmpCoords` were calculated in the `Atlas`, the user would have to calculate them himself, which cannot be the point.

I could imagine moving the calculation of the `BmpCoords` to a separate function `subCoords(x:Int, y:Int, w:Int, h:Int)` in `BmpCoordsWithT<T : ISizeInt>` in order to then it in the `sliceXXX()` functions. Maybe this reduces the complexity a bit since parentCoords are already known in this class. Would that be better?

Marking the property `rotated` (and also `rotatedAngle`) as deprecated makes sense. Or did you mean to completely omit the constructor with `rotated`?

Stechmuck:
@soywiz
I've moved most of the functionallity of the `init { }` method to `BmpCoordsWithT.subCoords()` and was able to remove some checks.

`basePixelPos()` now uses a lazy initialized private `PointInt` if not specified in the function call.

`rotated` and `rotatedAngle` are marked deprecated

korim/src/commonMain/kotlin/com/soywiz/korim/bitmap/BitmapSlice.kt
```diff
@@ -281,68 +393,194 @@ abstract class BmpSlice(
...
+            } else {
+                if (x < 0 || y < 0 || x >= frameWidth || y >= frameHeight) {
+                    throw IllegalArgumentException("Point $x,$y is not in bounds of slice")
+                }
+                bmpBase = extract()
+                bounds.setBoundsTo(0, 0, bmpBase.width, bmpBase.height)
+                virtFrame?.setBoundsTo(0, 0, bmpBase.width, bmpBase.height)
+                bmpCoords = BmpCoordsWithInstanceBase(
+                    SizeInt(bmpBase.width, bmpBase.height),
+                    0f, 0f, 1f, 0f, 1f, 1f, 0f, 1f
```
soywiz, line 504: What is this `0f, 0f, 1f, 0f, 1f, 1f, 0f, 1f`

korim/src/commonMain/kotlin/com/soywiz/korim/bitmap/BitmapSlice.kt
```diff
@@ -41,13 +47,82 @@ interface BmpCoordsWithT<T : ISizeInt> : BmpCoords, Closeable, Resourceable<BmpC
...
+            blX = brX
+            blY = tlY
+        } else {
+            // Base is not rotated
+            val xDim = if (imageOrientation.isRotatedDeg90CwOrCcw) h else w
+            val yDim = if (imageOrientation.isRotatedDeg90CwOrCcw) w else h
+            tlX = tl_x + x / bw * dx
+            tlY = tl_y + y / bh * dy
+            brX = tlX + xDim / bw * dx
+            brY = tlY + yDim / bh * dy
```
soywiz, line 116: These lines look duplicated (only nverting bw/bh)

soywiz:
@Stechmuck looks less scary now, there are some lines a bit duplicated and I'm not sure what the `0f, 0f, 1f, 0f, 1f, 1f, 0f, 1f` thing is but I have seen that in a few places. Maybe it could be the default or provide a constructor doing that?

korim/src/commonMain/kotlin/com/soywiz/korim/bitmap/BitmapSlice.kt
Stechmuck: `tl_x`, `tl_y`, `tr_x`, ... for a full size slice

korim/src/commonMain/kotlin/com/soywiz/korim/bitmap/BitmapSlice.kt
Stechmuck: Not really, all lines are different. `x`, `y`, `h` and `w` are also inverted and for example `trX` is assigned from `tlX` or `brX`

soywiz: I was thinking on something like this (untested, so not sure if I flipped something): 

```kotlin
val flipDim = isRotatedInBaseDeg90 xor imageOrientation.isRotatedDeg90CwOrCcw
val xDim = if (flipDim) h else w
val yDim = if (flipDim) w else h
val _bw = if (isRotatedInBaseDeg90) bh else bw
val _bh = if (isRotatedInBaseDeg90) bw else bh
val _x = if (isRotatedInBaseDeg90) y else x
val _y = if (isRotatedInBaseDeg90) x else y
tlX = tl_x + _x / _bw * dx
tlY = tl_y + _y / _bh * dy
brX = tlX + xDim / _bh * dx
brY = tlY + yDim / _bh * dy
```

But that way the arithmetic formulas are not repeated

korim/src/commonMain/kotlin/com/soywiz/korim/bitmap/BitmapSlice.kt
```diff
@@ -94,31 +81,22 @@ interface BmpCoordsWithT<T : ISizeInt> : BmpCoords, Closeable, Resourceable<BmpC
...
-            tlX = tl_x + x / bw * dx
-            tlY = tl_y + y / bh * dy
-            brX = tlX + xDim / bw * dx
-            brY = tlY + yDim / bh * dy
-            trX = brX
-            trY = tlY
-            blX = tlX
-            blY = brY
-        }
+        val flipDim = isRotatedInBaseDeg90 xor ioRotated
```
soywiz, line 84:
```suggestion
        val flipDim = isRotatedInBaseDeg90 xor isRotated
```

Stechmuck: io should stand for imageOrientation, but I have no Problem to apply this change

soywiz: Okay. Then lets keep as it is now. Ill trry to review it in detail later today and merge it if possible 👍

Stechmuck: Currently working on it 😄

soywiz:
@SchneckeddeF I have added the code initially provided by you in the original issue here: https://github.com/korlibs/korge-next/issues/622#issue-1225142733
And it is not displaying correctly for me:

You can try it yourself by executing: `./gradlew :korge-sandbox:runJvm`

Screenshot 2022-06-16 at 16 39 45

Stechmuck:
I'll have a look...

soywiz:
Also in the dropdown there is the `Dragonbones` sample that I just added to check since it also have support for rotated slices I guess:

It seems that something broke (maybe we need to adjust something on dragonbones side?):

Screenshot 2022-06-16 at 16 43 11

Stechmuck:
I have updated RotatedAtlas with an actual implementation.

Can you pleas give me a hint, where the Dragonbone Atlas is parsed. I think there have to be made some changes as I made in SpineRegion

soywiz-invideo:
@Stechmuck It has been a while, but my bet goes here:
* https://github.com/korlibs/korge-next/blob/main/korge-dragonbones/src/commonMain/kotlin/com/dragonbones/model/TextureAtlasData.kt
* https://github.com/korlibs/korge-next/blob/main/korge-dragonbones/src/commonMain/kotlin/com/soywiz/korge/dragonbones/KorgeDbTextureAtlasData.kt

https://github.com/korlibs/korge-next/blob/76f1835179e29e64c6adf58f0720d9b261dbd615/korge-dragonbones/src/commonMain/kotlin/com/soywiz/korge/dragonbones/KorgeDbTextureAtlasData.kt#L86

Would be nice to understand the kind of breaking change that is causing this and either put a compatibility something or at least know about it to explain it in the migration guide from KorGE 2.0 -> KorGE 3.0

Stechmuck:
@soywiz

Should work now as expected. It was a missing rotation support in the deprecated constructors.

soywiz-invideo:
@Stechmuck Cool. I'll check later everything and in internal use-cases that uses texture slicing/transformations. Any chance to get a regression test to avoid that issue from happening with those constructors? Even if they are deprecated constructors, might be useful

soywiz:
@Stechmuck I have tried and couldn't find any major issue with the code I tried. Just merged it.

Thank you very much for reporting the bug, actually fixing it, doing a great job and your huge patience. It has been almost a whole month!

PR #645, opened 2022-05-13
Add ByteArray2, CharArray2, BooleanArray2, LongArray2, make AStar implementation to use BooleanArray2, split in files, remove redundant `: Unit` return type and optimize imports to reduce compile times

Kietyo:
- Implemented ByteArray2, CharArray2, BooleanArray2, LongArray2.
- Some simple auto fixes from Intellij.
- Many imports were using wildcards, which increase compile time. Intellij is able to auto optimize those imports.


kds/src/commonTest/kotlin/com/soywiz/kds/Array2Test.kt
```diff
@@ -85,4 +86,33 @@ class Array2Test {
...
+        }
+        val doubleArray2 = DoubleArray2(2, 2) {
+            it.toDouble()
+        }
+        val typedArray2 = Array2<Boolean>(2, 2) {
+            (it % 2) == 1
+        }
+
+        intArray2.each { x, y, v ->
+            println("x: $x, y: $y, v: $v")
```
soywiz, line 106: This is printing but not asserting anything, right?

kds/src/commonMain/kotlin/com/soywiz/kds/BooleanArray2.kt
```diff
@@ -0,0 +1,107 @@
+package com.soywiz.kds
+
+@Suppress("NOTHING_TO_INLINE", "RemoveExplicitTypeArguments")
+data class BooleanArray2(override val width: Int, override val height: Int, val data: BooleanArray) :
```
soywiz, line 4: This partially covers https://github.com/korlibs/kds/issues/1

korma/src/commonMain/kotlin/com/soywiz/korma/algo/AStar.kt
```diff
@@ -1,13 +1,16 @@
...
+import com.soywiz.kds.BooleanArray2
+import com.soywiz.kds.IntPriorityQueue
+import com.soywiz.korma.geom.IPointIntArrayList
+import com.soywiz.korma.geom.Point
+import com.soywiz.korma.geom.PointIntArrayList
 
 class AStar(val width: Int, val height: Int, val isBlocking: (x: Int, y: Int) -> Boolean) {
     companion object {
-        operator fun invoke(board: Array2<Boolean>) = AStar(board.width, board.height) { x, y -> board[x, y] }
+        operator fun invoke(board: BooleanArray2) = AStar(board.width, board.height) { x, y -> board[x, y] }
```
soywiz, line 11: This is changed. Shouldn't we keep the also the old constructors?

korau/src/commonMain/kotlin/com/soywiz/korau/format/mp3/minimp3/MiniMp3Program.kt
```diff
@@ -90,7 +90,7 @@ internal open class MiniMp3Program(HEAP_SIZE: Int = 0) : Runtime(HEAP_SIZE) {
     // typealias L12_subband_alloc_t = L12_subband_alloc_t
     // typealias L3_gr_info_t = L3_gr_info_t
     // typealias mp3dec_scratch_t = mp3dec_scratch_t
-    fun bs_init(bs: CPointer<bs_t>, data: CPointer<UByte>, bytes: Int): Unit {
+    fun bs_init(bs: CPointer<bs_t>, data: CPointer<UByte>, bytes: Int) {
```
soywiz, line 93: Note, since this code is autogenerated and I might need to update it in the future when further optimizations are done. We might want to update also here:
https://github.com/korlibs/ktcc/blob/f3d1cace49f6a74e3aad71a2cdcc22b877d3c783/src/commonMain/kotlin/com/soywiz/ktcc/gen/Kotlin.kt#L380

korma/src/commonTest/kotlin/com/soywiz/korma/algo/AStarTest.kt
```diff
@@ -63,24 +67,24 @@ class AStarTest {
     }
 
 
-    data class Result(val map: Array2<Boolean>, val start: IPointInt, val end: IPointInt)
+    data class Result(val map: BooleanArray2, val start: IPointInt, val end: IPointInt)
```
soywiz, line 70: Oh, also changed here

soywiz:
So I guess the point here was optimizing the AStar implementation right?

This codebase is indeed a mess in terms of linting, so it might make sense to add a linter already and do an initial reformating so we have a source of truth in terms of formatting.

I have noticed that some imports were still keept as `*` in this PR. Sometime ago I believe the android rules were no * at all, and the Kotlin typical rules were * everywhere. Though probably resolving * requires more time. Though that would make moving some methods to extensions a breaking source-code change, which is a bit unfortunate. But well.

I believe we should also split PRs: adding functionality, refactors and linting stuff, and I should be the first one to follow that :) 

Since this is likely to produce conflicts later, I'm going to merge it already since tests are passing, but we should cover at least the println in tests not doing assertions later

Kietyo:
Actually I was just bored so decided to implement some random stuff while listening to podcasts.

kds/src/commonTest/kotlin/com/soywiz/kds/Array2Test.kt
Kietyo: I was having some issues in my intellij where it wasn't seeing the `.each` methods, but rebuilding the project fixed it. So I guess these tests can be removed.

soywiz:
Cool, that's always a nice thing to do :)

Kietyo:
Note that some imports still has * because I ran the "auto optimize import" before updating these settings:

image

I'll be sending out a few PR to optimize modules in the next few days.

PR #700, opened 2022-05-29
More versatile function to create rectangle from 2 points.

Kietyo:


korma/src/commonMain/kotlin/com/soywiz/korma/geom/IRectangle.kt
```diff
@@ -0,0 +1,92 @@
...
+        //
+        // Rect XY   Point 2
+        // │        │
+        // ▼        ▼
+        // ┌────────┐
+        // │        │
+        // │        │
+        // └────────┘
+        // ▲
+        // │
```
soywiz, line 34: This is a bit misleading, since it says the top left point is 2, but in the diagram it is the bottom left one, and in the implementation it doesnt seems to matter since in the end it uses min and max functions

korma/src/commonMain/kotlin/com/soywiz/korma/geom/IRectangle.kt
```diff
@@ -0,0 +1,92 @@
...
+        }
+    }
+}
+
+@Deprecated(
+    "Properties with underscores are deprecated and will be removed soon",
+    ReplaceWith("x"),
+    DeprecationLevel.ERROR
+)
+val IRectangle._x get() = x
```
soywiz, line 51: Yeah, we should even remove them already since we are going into 3.0 and this has been here for quite a long time already

korma/src/commonMain/kotlin/com/soywiz/korma/geom/IRectangle.kt
```diff
@@ -0,0 +1,92 @@
...
+    DeprecationLevel.ERROR
+)
+val IRectangle._height get() = height
+
+val IRectangle.left get() = x
+val IRectangle.top get() = y
+val IRectangle.right get() = x + width
+val IRectangle.bottom get() = y + height
+
+val IRectangle.topLeft get() = Point(left, top)
```
soywiz, line 79: This is allocating. Until we have proper inline classes for multiple args, in the APIs we try to use methods with a final argument called out with a default constructor to use them as a holder to avoid allocations. Then there is a PointPool that can have extension methods that avoid allocations. It is a bit painful but a way yo avoid allocation

Kietyo: BTW I didnt change this code. Simply moved it to another file. Feel free to submit another PR for that stuff.

korma/src/commonMain/kotlin/com/soywiz/korma/geom/IRectangle.kt
Kietyo: The order or position of the points does not matter, clarified further.

korma/src/commonMain/kotlin/com/soywiz/korma/geom/IRectangle.kt
```diff
@@ -0,0 +1,107 @@
...
+        // └────────┘
+        //          ▲
+        //          │
+        //        point2
+        operator fun invoke(point1: IPoint, point2: IPoint): Rectangle {
+            val topLeftX = minOf(point1.x, point2.x)
+            val topLeftY = minOf(point1.y, point2.y)
+            val bottomRightX = maxOf(point1.x, point2.x)
+            val bottomRightY = maxOf(point1.y, point2.y)
+            return Rectangle(topLeftX, topLeftY, bottomRightX - topLeftX, bottomRightY - topLeftY)
```
soywiz, line 56: What about renaming:

```suggestion
            val left = minOf(point1.x, point2.x)
            val top = minOf(point1.y, point2.y)
            val right = maxOf(point1.x, point2.x)
            val bottom = maxOf(point1.y, point2.y)
            return Rectangle(left, top, right - left, bottom - top)
```

So it is simpler and clearer?

soywiz:
Thanks!

PR #712, opened 2022-05-30
Implements the function that changes the length of a Point

mariorez:
I'm developing a game where I change the speed of the object by changing the length (magnitude) of the Point:
https://github.com/mariorez/korge-starfish-collector/blob/main/src/commonMain/kotlin/system/MovementSystem.kt#L46
https://github.com/mariorez/korge-starfish-collector/blob/main/src/commonMain/kotlin/system/MovementSystem.kt#L76

Here is my proposal to add the functionality of `setLength` in the `Point` class to meet this use case.


mariorez:
This implementation is a port of LibGDX's `setLength` method:
https://github.com/libgdx/libgdx/blob/master/gdx/src/com/badlogic/gdx/math/Vector2.java#L259

korma/src/commonMain/kotlin/com/soywiz/korma/geom/Point.kt
```diff
@@ -317,6 +317,17 @@ data class Point(
 
     fun rotate(rotation: Angle, out: Point = Point()): Point =
         out.setToPolar(Angle.between(0.0, 0.0, this.x, this.y) + rotation, this.length)
+
+    fun setLength(length: Double): Point {
```
soywiz, line 321: I'm not sure about the name, because set methods usually do not depends on internal state.

I would propose other options:

Allowing to set everything (but defaulting to the internal value), similar to setPolar:
```suggestion
    fun setLength(length: Double, x: Double = this.x, y: double = this.y): Point {
```

```suggestion
    fun setLengthKeepingAngle(length: Double): Point {
```

```suggestion
    fun changeLength(length: Double): Point {
```

mariorez: I believe the `changeLength` option is the best, as it clearly describes the intent of this method.
I will make the adjustments and submit for your review.

mariorez:
I realized that the `changeLength` method is not accurate in all scenarios:
https://github.com/korlibs/korge-next/pull/712/files#diff-8413bc61de5cb0946163fc468a74c95a953727168a2a228a65b02c9fa7f47c8cR55

Could this be a problem?

mariorez:
the new calculation approach also does not guarantee accurate results 😞 
**Implementation**: https://github.com/korlibs/korge-next/pull/712/files#diff-79dadca66643236a9d53bfacc28326eeb175e669a023dd4ffb03176e9f155105R323
**Test**: https://github.com/korlibs/korge-next/pull/712/files#diff-8413bc61de5cb0946163fc468a74c95a953727168a2a228a65b02c9fa7f47c8cR55

```
this.normalize()
this.setTo(this.x * newLength, this.y * newLength)
```

korma/src/commonMain/kotlin/com/soywiz/korma/geom/Point.kt
```diff
@@ -317,6 +317,21 @@ data class Point(
...
+    fun changeLength(length: Int): Point = changeLength(length.toDouble())
+    fun changeLength(length: Double): Point {
+        this.normalize()
+        this.setTo(this.x * length, this.y * length)
+        return this
+    }
+
+    val angle: Angle
+        get() {
+            var angle: Double = atan2(this.y, this.x) * RAD2DEG
```
soywiz, line 331: https://github.com/korlibs/korge-next/blob/61d5d25b2a486338f913d9c8847850587c599f76/korma/src/commonMain/kotlin/com/soywiz/korma/geom/Angle.kt#L78

korma/src/commonTest/kotlin/com/soywiz/korma/geom/PointTest.kt
```diff
@@ -14,4 +14,57 @@ class PointTest {
...
+        assertEquals(45.degrees, point.angle)
+    }
+
+    @Test
+    fun testChangeLength_ISNOT_accurate_in_all_scenarios() {
+        val point = Point(5, 5)
+        assertEquals(45.degrees, point.angle)
+
+        point.changeLength(30.0)
+        assertEquals(29.999999999999996, point.length)
```
soywiz, line 55: well in any case  you can use the third parameter in `assertEquals` to compare floating point values to set the epsilon value so you can use assertEquals(30.0, point.length, 0.01)

korma/src/commonMain/kotlin/com/soywiz/korma/geom/Point.kt
mariorez: Done 👍

korma/src/commonTest/kotlin/com/soywiz/korma/geom/PointTest.kt
mariorez: Good idea and I believe it is a good solution for this scenario:
https://github.com/korlibs/korge-next/pull/712/files#diff-8413bc61de5cb0946163fc468a74c95a953727168a2a228a65b02c9fa7f47c8cR55

korma/src/commonMain/kotlin/com/soywiz/korma/geom/Point.kt
```diff
@@ -321,14 +321,15 @@ data class Point(
...
         return this
     }
 
     val angle: Angle
         get() {
-            var angle: Double = atan2(this.y, this.x) * RAD2DEG
+            var angle = atan2(this.y, this.x).degrees
             if (angle < 0) angle += 360.0
             return angle.degrees
         }
```
soywiz, line 335: Actually, you can just do:

```kotlin
return Angle.atan2(this.y, this.x)
```
or
```kotlin
return Angle.between(0.0, 0.0, this.x, this.y).normalized // Normalizing is converting any angle into the right 0..360 degrees
```


Why is the x and y in opposite places, and why are you normalizing the angle later? Is this what people would expect? Shouldn't it be something like `normalizedAngle`? Or just provide the angle and let people normalize it if they need it?

PR #715, opened 2022-06-01
Deprecate global injection object for korge fleks for world local injections.

Kietyo:
The idea is that each world should have their own injection space. Currently, every world that's created overwrites the previous world's injections. So you can't have 2 worlds with different injections.

This is useful for simulation/multiplayer type games where each player has their own world.

A real life example of this is that for my game, I am replicating a starcraft 2 arcade map called "Entropy TD" where each player is competing against each other with their own mazes. Each player has their own world with their own instances of injectables.

https://sc2arcade.com/map/2/148730/


Kietyo:
Thanks, I'll try it out tomorrow.

PR #724, opened 2022-06-06
fix NPE in PolygonShape.set when no Vec2ArrayPool is given

jbellis:
Note that the cast on line 133 is necessary for Kotlin to figure out that the actual fix on line 147 is valid

### DIFF
--- a/kbox2d/src/commonMain/kotlin/org/jbox2d/collision/shapes/PolygonShape.kt
+++ b/kbox2d/src/commonMain/kotlin/org/jbox2d/collision/shapes/PolygonShape.kt
@@ -128,10 +128,11 @@ class PolygonShape : Shape(ShapeType.POLYGON) {
         var n = MathUtils.min(num, Settings.maxPolygonVertices)
 
         // Perform welding and copy vertices into local buffer.
-        val ps = if (vecPool != null)
-            vecPool[Settings.maxPolygonVertices]
+        val ps : Array<Vec2?>
+        if (vecPool != null)
+            ps = vecPool[Settings.maxPolygonVertices] as Array<Vec2?>
         else
-            arrayOfNulls<Vec2>(Settings.maxPolygonVertices)
+            ps = arrayOfNulls<Vec2>(Settings.maxPolygonVertices)
         var tempCount = 0
         for (i in 0 until n) {
             val v = verts[i]
@@ -144,7 +145,7 @@ class PolygonShape : Shape(ShapeType.POLYGON) {
             }
 
             if (unique) {
-                ps[tempCount++]!!.set(v)
+                ps[tempCount++] = v
             }
         }
 
@@ -212,9 +213,6 @@ class PolygonShape : Shape(ShapeType.POLYGON) {
 
         // Copy vertices.
         for (i in 0 until count) {
-            if (vertices[i] == null) {
-                vertices[i] = Vec2()
-            }
             vertices[i].set(ps[hull[i]]!!)
         }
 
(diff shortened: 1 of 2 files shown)


kbox2d/src/commonMain/kotlin/org/jbox2d/collision/shapes/PolygonShape.kt
```diff
@@ -212,9 +213,6 @@ class PolygonShape : Shape(ShapeType.POLYGON) {
 
         // Copy vertices.
         for (i in 0 until count) {
-            if (vertices[i] == null) {
-                vertices[i] = Vec2()
-            }
```
soywiz, line 217: Why is this removed?

kbox2d/src/commonMain/kotlin/org/jbox2d/collision/shapes/PolygonShape.kt
```diff
@@ -128,10 +128,11 @@ class PolygonShape : Shape(ShapeType.POLYGON) {
...
 
         // Perform welding and copy vertices into local buffer.
-        val ps = if (vecPool != null)
-            vecPool[Settings.maxPolygonVertices]
+        val ps : Array<Vec2?>
+        if (vecPool != null)
+            ps = vecPool[Settings.maxPolygonVertices] as Array<Vec2?>
         else
-            arrayOfNulls<Vec2>(Settings.maxPolygonVertices)
+            ps = arrayOfNulls<Vec2>(Settings.maxPolygonVertices)
```
soywiz, line 135: This change shouldn't be required, right?

soywiz:
Can you also add your failing test case here?
https://github.com/korlibs/korge-next/tree/f01a13a932e4ed4f96c39a20604df902ebf7ef38/kbox2d/src/commonTest/kotlin/org/jbox2d/utests

Something like:

```kotlin
class PolygonShapeTest {
  @Test
  fun testPolygonShapeTest() {
    val shape = PolygonShape()
    var zero = Vec2(0f, 0f)
    shape.set(arrayOf(zero, zero, zero), 3) // This shouldn't throw
    assertEquals(...)
  }
}
```

soywiz:
Fixes https://github.com/korlibs/korge/issues/445

kbox2d/src/commonMain/kotlin/org/jbox2d/collision/shapes/PolygonShape.kt
jbellis: Above: 

Note that the cast on line 133 is necessary for Kotlin to figure out that the actual fix on line 147 is valid

jbellis:
test added to PR

kbox2d/src/commonMain/kotlin/org/jbox2d/collision/shapes/PolygonShape.kt
soywiz: What makes you think it is unnecessary in all the cases?

kbox2d/src/commonTest/kotlin/org/jbox2d/utests/PolygonShapeTest.kt
```diff
@@ -0,0 +1,23 @@
...
+
+
+class PolygonShapeTest {
+    @Test
+    fun test() {
+        val shape = PolygonShape()
+        val vertexArray = arrayOf(Vec2(1f, 1f), Vec2(1f, 0f), Vec2(0f, 0f))
+        shape.set(vertexArray, vertexArray.size)
+
+        assertEquals(shape.count, vertexArray.size)
```
soywiz, line 18: Cool. Thanks!

kbox2d/src/commonMain/kotlin/org/jbox2d/collision/shapes/PolygonShape.kt
soywiz: Now the declaration and the assignment is split but it looks the same to me and it was compiling before. What happens if you restore the original lines there?

soywiz:
Just checked that vectors: Array<Vec2> is already initialized, so indeed not needed. Going to merge it now.
Thanks!

PR #731, opened 2022-06-10
Add JetPack compose support to KorGE (Composable)

soywiz:
Since compose runtime is not available on all the korge K/N targets (for example linux, mingw, ios, etc.), I have included the source of the compatible runtime with this compose version. Also it doens't work with Kotlin 1.7.0 yet.

This proposal is for using Composable functionality inside KorGE. The idea here is that you can for example create the gameplay of a complex game using normal mutable functionality, while create the UI and things like that using composables, which is more suitable for that kind of work.

* [x] Integration with Composable Runtime (unfortunately had to copy it here to support some targets, since published runtime seems to be only for JVM, JS, Android and MacosArm64, and we are lacking the rest of K/N targets)
* [ ] Integratiom with Composable Platform / UI
* [x] Partial mimicking some typical composable functionality, like basic layouts, components and `Modifier`


soywiz:
Woops this was closed automatically

soywiz:
Moved here: https://github.com/korlibs/korge/pull/485

soywiz:
Moved here: https://github.com/korlibs/korge/pull/485

PR #763, opened 2022-06-20
Introduce UI Dropdown. Update Sandbox to use it.

Kietyo:
I found the combobox to be a bit finnicky. Brought in the dropdown from my own project for use.


soywiz:
Hey @Kietyo thanks for the PR!
Unfortunately I cannot use this component as it is right now. The UIComboBox is designed to work on mobile too (with finger scroll + intertia) that helps testing the demos on Android and iOS where you don't have a mouse to do scrolling.

I have increased the size of the dropdown, and sorted the demos by name. 
Which aspects do you believe UIComboBox should improve?

